PR #4, opened 2019-04-02
Zcu102 hyp

chrisguikema:
These commits add platform support for the zynqmp. They also fix a bug that caused RAM regions to be truncated, which broke the zynqmp since its loaded into high memory.

### DIFF
--- a/libsel4arm-vmm/src/devices.c
+++ b/libsel4arm-vmm/src/devices.c
@@ -227,7 +227,7 @@ void *map_emulated_device(vm_t *vm, const struct device *d)
     /* Map the frame to the VM */
     DMAP("Mapping emulated device ipa0x%x\n", (uint32_t)vm_addr);
 
-    seL4_CapRights_t rights = seL4_CanRead;
+    seL4_CapRights_t rights = seL4_NoRights;
     res = vspace_reserve_range_at(vm_vspace, vm_addr, size, rights, 0);
     assert(res.res);
     if (!res.res) {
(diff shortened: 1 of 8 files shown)


WellMcGarnicle:
Cool, this looks good :+1:  I'll hold off on merging it until the kernel changes are ready

chrisguikema:
The kernel changes have been merged!

libsel4arm-vmm/src/devices.c
```diff
@@ -227,7 +227,7 @@ void *map_emulated_device(vm_t *vm, const struct device *d)
     /* Map the frame to the VM */
     DMAP("Mapping emulated device ipa0x%x\n", (uint32_t)vm_addr);
 
-    seL4_CapRights_t rights = seL4_CanRead;
+    seL4_CapRights_t rights = seL4_NoRights;
```
kent-mcleod: Are you happy for me to revert this?

chrisguikema: Why should an emulated device have Read Rights?

We've emulated the serial device for the zynqmp, and when there are read rights, emulating doesn't work.

kent-mcleod: Creating an emulated device without read or write rights doesn't require any frames to be mapped into the vm and this can be implemented with:
```c
-    vgic->dist = map_emulated_device(vm, &dev_vgic_dist);
+    vgic->dist = malloc(dev_vgic_dist.size);
```

chrisguikema:
Created a new PR since this one is so outdated...

https://github.com/seL4/seL4_projects_libs/pull/78

PR #5, opened 2019-07-31
Static lr of

pingerino:


libsel4arm-vmm/src/arch/arm/devices/vgic.c
```diff
@@ -292,21 +297,11 @@ static int vgic_vcpu_inject_irq(struct device *d, vm_t *vm, struct virq_handle *
...
-        list_size++;
-        lrof = (struct lr_of *)malloc(sizeof(*lrof));
-        assert(lrof);
-        if (lrof == NULL) {
-            return -1;
-        }
-        lrof->virq_data = irq;
-        lrof->next = NULL;
-        *lrof_ptr = lrof;
+        ZF_LOGF_IF(vgic->lr_overflow.head == vgic->lr_overflow.tail, "No overflow irqs to service");
```
WellMcGarnicle: Slightly confused about this check. Won't this condition be hit the first time we overflow? (from the call path `vm_inject_IRQ` -> `vgic_dist_set_pending_irq` -> `vgic_vcpu_inject_irq`)

pingerino: yup that makes no sense, will remove

PR #6, opened 2019-08-01
vgic cleanup to prep for gicv3

pingerino:


pingerino:
I'm going to close this for now since I've ended up reworking a fair bit. Consider it a preview :).

PR #7, opened 2019-08-02
libsel4arm-vmm: fix bugs in ba8de51828024fe1de88b2

pingerino:
- out-by-1 in LR_OF_NEXT
  - bad assertion in vgic_vcpu_inject_irq
  - inifinite loop in handle_vgic_maintenance

  handle_vgic_maintenance now processes as many overflow irqs as existed
  at the start of the procedure call.

### DIFF
--- a/libsel4arm-vmm/src/arch/arm/devices/vgic.c
+++ b/libsel4arm-vmm/src/arch/arm/devices/vgic.c
@@ -158,6 +158,7 @@ struct lr_of {
     struct virq_handle irqs[MAX_LR_OVERFLOW]; /* circular buffer */
     size_t head;
     size_t tail;
+    bool full;
 };
 
 typedef struct vgic {
@@ -200,6 +201,7 @@ static int virq_init(vgic_t *vgic)
     memset(vgic->virqs, 0, sizeof(vgic->virqs));
     vgic->lr_overflow.head = 0;
     vgic->lr_overflow.tail = 0;
+    vgic->lr_overflow.full = false;
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


libsel4arm-vmm/src/arch/arm/devices/vgic.c
```diff
@@ -297,11 +297,11 @@ static int vgic_vcpu_inject_irq(struct device *d, vm_t *vm, struct virq_handle *
         return err;
     } else {
         /* Add to overflow list */
-        int idx = LR_OF_NEXT(vgic->lr_overflow.tail);
-        ZF_LOGF_IF(idx == vgic->lr_overflow.head && idx != vgic->lr_overflow.tail,
-                   "too many overflow irqs");
+        int idx = vgic->lr_overflow.tail;
+        ZF_LOGF_IF(idx == vgic->lr_overflow.head &&
+                   vgic->lr_overflow.head != vgic->lr_overflow.tail, "too many overflow irqs");
```
WellMcGarnicle: Why is it an error if the head does not equal the tail? Also if the first condition is true the second must be false?

pingerino: When `head == tail` the buffer is empty, so we can allocated the head in this case.

pingerino: otherwise the buffer is full and we're about to overwrite the current head

pingerino: Thanks for the fix!

PR #16, opened 2021-03-26
Dornerworks x64 VM patch

nomadeel:
This PR contains commits from Dornerworks' x64 VM patches for the VMM: https://github.com/dornerworks/camkes-vm/commits/open_source_release. Parts of Dornerworks' patches have been updated and/or changed slightly to match the newer VMM structure compared to what the patches originally targeted.


oliver-wm:
Is this the one that was on bitbucket?

nomadeel:
I think so, yeah.

oliver-wm:
> I think so, yeah.
> 
> EDIT: No, nevermind, it's not. This is new.

Ok will review then

libsel4vm/src/arch/x86/vmcs.c
```diff
@@ -62,8 +97,13 @@ void vm_vmcs_init_guest(vm_vcpu_t *vcpu)
     vm_vmcs_write(vcpu->vcpu.cptr, VMX_GUEST_RFLAGS, BIT(1));
     vm_vmcs_write(vcpu->vcpu.cptr, VMX_GUEST_SYSENTER_ESP, 0);
     vm_vmcs_write(vcpu->vcpu.cptr, VMX_GUEST_SYSENTER_EIP, 0);
-    vcpu->vcpu_arch.guest_state->machine.control_ppc = VMX_CONTROL_PPC_HLT_EXITING | VMX_CONTROL_PPC_CR3_LOAD_EXITING |
-                                                       VMX_CONTROL_PPC_CR3_STORE_EXITING;
+
+#ifdef CONFIG_ARCH_X86_64
+    vm_vmcs_init_64_bit_guest(vcpu);
+#else
+    vm_vmcs_init_32_bit_guest(vcpu);
```
kent-mcleod, line 104: There are example apps that run a 32bit guest on a 64 bit host.  Is this no longer something that we will support?

kent-mcleod (approved):

libsel4vm/src/arch/x86/vmcs.c
nomadeel: Yep, just found out as well when fixing some builds. I need to make extra changes to the various repositories to allow 32-bit guest on 64-bit host.

wom-bat: Have these changes been made?

libsel4vm/src/arch/x86/ept.c
```diff
@@ -28,13 +28,13 @@
...
     /* Read linear address that guest is trying to access. */
-    unsigned int linear_address;
+    seL4_Word linear_address;
     vm_vmcs_read(vcpu->vcpu.cptr, VMX_DATA_GUEST_LINEAR_ADDRESS, &linear_address);
     printf(COLOUR_R "!!!!!!!! ALERT :: GUEST OS PAGE FAULT !!!!!!!!\n");
     printf("    Guest OS VMExit due to EPT Violation:\n");
-    printf("        Linear address 0x%x.\n", linear_address);
-    printf("        Guest-Physical address 0x%x.\n", vm_guest_exit_get_physical(vcpu->vcpu_arch.guest_state));
-    printf("        Instruction pointer 0x%x.\n", vm_guest_state_get_eip(vcpu->vcpu_arch.guest_state));
+    printf("        Linear address 0x%lx.\n", linear_address);
```
axel-h, line 35: The proper way to print a 'seL4_Word' would be using the dedicated `SEL4_PRIx_word, that adapts to the arch/plat specific format. Please don't use "lx".

chrisguikema:
What needs to happen to get this merged?

libsel4vm/src/arch/x86/guest_x86_context.c
```diff
@@ -4,6 +4,13 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 
+/* Under release mode, this file will get built using -O3. Howerver on gcc 8.4,
```
axel-h, line 7: Is this a problem specific to GCC 8.4 or does it happen with newer versions and clang also? Such issues leave an odd feeling that there is an issue with the code (maybe alignment?) that I consider worth investigating. However, it should not be seen as a blocker for merging this as it temporarily fixes a problem.

I have created https://github.com/seL4/seL4_projects_libs/issues/75 to track this.

axel-h:
> What needs to happen to get this merged?

Seems it needs a rebase now. I have a few comments on the code quality, but have to leave it to other to comment on the x64 changes.

libsel4vm/arch_include/x86/sel4vm/arch/vmcs_fields.h
```diff
@@ -95,3 +95,23 @@
...
+#define ACCESS_RIGHTS_L        BIT(13)
+#define ACCESS_RIGHTS_DB       BIT(14)
+#define ACCESS_RIGHTS_G        BIT(15)
+#define ACCESS_RIGHTS_UNUSABLE BIT(16)
+
+#define ACCESS_RIGHTS_TYPE_BUSY_TSS    0xB
+#define ACCESS_RIGHTS_TYPE_BUSY_16_TSS 0x3
+#define ACCESS_RIGHTS_TYPE_LDT         0x2
+
+#define DEFAULT_TR_ACCESS_RIGHTS AR(P) | AR_T(BUSY_TSS)
```
axel-h, line 116: Please use brackets here and in the other macro to make the expression an atoms.

libsel4vm/src/arch/x86/boot.c
```diff
@@ -76,25 +83,119 @@ static vm_frame_t pd_alloc_iterator(uintptr_t addr, void *cookie)
...
+static int make_guest_pd_continued(void *access_addr, void *vaddr, void *cookie)
+{
+    uint64_t *pd = vaddr;
+    int num_entries = BIT(seL4_PageBits) / sizeof(pd[0]);
 
-static int make_guest_page_dir(vm_t *vm)
+    /* Brute force 1:1 entries. */
+    for (int i = 0; i < num_entries; i++) {
+        /* Present, write, user, page size 2M */
+        pd[i] = (uint64_t)((uint64_t)((i) << PAGE_BITS_2M) | PAGE_ENTRY);
```
axel-h, line 94: Better cast the base type to `uint64_t`, then all operation will happen on an integer of this type
```suggestion
        pd[i] = ((uint64_t)i) << PAGE_BITS_2M) | PAGE_ENTRY;
```

libsel4vm/src/arch/x86/processor/decode.c
```diff
@@ -368,15 +499,65 @@ uintptr_t vm_emulate_realmode(vm_vcpu_t *vcpu, uint8_t *instr_buf,
             case 0xa1:
                 /* mov offset memory to eax */
                 instr++;
+#ifdef CONFIG_ARCH_X86_64
+                memcpy(&mem, instr, 4);
```
axel-h, line 503: When using `sizeof(seL4_Word)/2` there is no need to duplicate the code. Might apply in other places also.
```suggestion
                memcpy(&mem, instr, sizeof(seL4_Word)/2 );
```

libsel4vmmplatsupport/src/arch/x86/acpi.c
```diff
@@ -257,16 +257,19 @@ int make_guest_acpi_tables(vm_t *vm)
...
     acpi_fill_table_head(&xsdt->header, "XSDT", 1);
 
     // Add previous tables to XSDT pointer list
     uintptr_t table_paddr = xsdt_addr + xsdt_size;
+    uintptr_t table_vaddr = xsdt_vaddr + xsdt_size;
     uint64_t *entry = (uint64_t *)((char *)xsdt + sizeof(acpi_xsdt_t));
     for (int i = 1; i < num_tables; i++) {
-        *entry++ = (uint64_t)table_paddr;
+        *entry++ = (uint64_t)table_vaddr;
         table_paddr += table_sizes[i];
```
axel-h, line 271: Seems the value of `table_paddr` is no longer used here? So the initialization and update could be removed for this loop.

libsel4vm/src/arch/x86/processor/msr.c
```diff
@@ -20,17 +20,44 @@
...
+
+static void vm_msr_write(seL4_CPtr vcpu, unsigned int field, seL4_Word value)
+{
+
+    seL4_X86_VCPU_WriteMSR_t result;
+
+    assert(vcpu);
+
+    result = seL4_X86_VCPU_WriteMSR(vcpu, (seL4_Word)field, value);
+    assert(result.error == seL4_NoError);
```
axel-h, line 45: Doesn't this complain that the value assigned to `result` is unused in release builds?
Consider using this:
```
result = ...
ZF_LOGF_IF(result.error != result, "writing failes, result = ...");
```
which always calls `abort()` then the condition fails. Otherwise declare declare `seL4_X86_VCPU_WriteMSR_t UNUSED result;` or use something like this:
```
if (result.error != result {
    ZF_LOGE("writing failed, result = ...");
    assert(0);
```

abrandnewusername:
Update: 
- Fixed most of the issues mentioned in previous comments
- Changed the guard from CONFIG_ARCH_X86_64 to CONFIG_X86_64_VTX_64BIT_GUESTS due to a decision made in the related PR: https://github.com/seL4/seL4/pull/324#discussion_r1180140561).

Future works: Current implementation for x64 support relies on the corresponding kernel implementation (see https://github.com/seL4/seL4/pull/324), which does not support running both 32-bit VM guest and 64-bit VM guest on the same system. It will be ideal to implement this feature in the future.

wom-bat (approved):
A few minor changes can clean up some of the ifdefs.  Otherwise, good to go.

PR #21, opened 2021-09-02
vgic.h: global should be declared extern

lsf37:
gcc before version 10 silently merged these, from v10 on the linker fails with multiply defined symbols. We could pass `-fcommon` to get the old gcc behaviour, but it's better to fix the code.

### DIFF
--- a/libsel4vm/src/arch/arm/vgic/vgic.h
+++ b/libsel4vm/src/arch/arm/vgic/vgic.h
@@ -12,7 +12,7 @@ struct vgic_dist_device {
     void *priv;
 };
 
-const struct vgic_dist_device dev_vgic_dist;
+extern const struct vgic_dist_device dev_vgic_dist;
 
 int vm_install_vgic(vm_t *vm);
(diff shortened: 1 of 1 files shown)


kent-mcleod (approved):

axel-h:
It's too late now, but I'm a bit unhappy about the commit comment suggesting "-fcommon" given we have https://github.com/seL4/seL4/commit/9c1508d8d97934598c23f4e5b361337fa06c2c20 that enforces "fno-common". Isn't declaring (ie. instanciating) variables in header files just a bad practice in the end?

lsf37:
Yes absolutely, that's what the commit message says, no?

The  commit doesn't do `-fcommon`, it fixes the source instead. It also doesn't suggest to use `-fcommon`, it only mentions it to indicate what changed in the compiler.

axel-h:
Yes, all fine. In the end I'm maybe just surprised once more how and why C works. And how naive (but nice) my view was before reading into the "common" stuff.

lsf37:
Yes, the deeper one goes the more frightening it gets..

PR #25, opened 2022-02-03
libsel4vmmplatsupport: fix PCI passthrough on ARM

hlyytine:
libsel4vmmplatsupport: fix PCI passthrough on ARM
    
BAR size resolution was emulated in two places, the fault handler and
the backend created with vmm_pci_create_passthrough_bar_emulation().
Move to vmm_pci_create_bar_emulation() on ARM as well and remove the
former. This way there is no need for BAR hack in the fault handler and
the backend created with vmm_pci_create_passthrough() works now on ARM.
    
Fix also several access size and bounds issues.

Signed-off-by: Hannu Lyytinen <hannux@ssrc.tii.ae>

Test with: seL4/camkes-vm#18

### DIFF
--- a/libsel4vmmplatsupport/src/arch/arm/devices/vpci.c
+++ b/libsel4vmmplatsupport/src/arch/arm/devices/vpci.c
@@ -271,21 +271,27 @@ int fdt_generate_vpci_node(vm_t *vm, vmm_pci_space_t *pci, void *fdt, int gic_ph
     bool is_irq_map = false;
     /* The first device is always the bridge (which doesn't need to be recorded in the ranges) */
     for (int i = 1; i < 32; i++) {
-        if (pci->bus0[i][0]) {
-            pci_bar_emulation_t *bar_emul = (pci_bar_emulation_t *)(pci->bus0[i][0])->cookie;
-            vmm_pci_entry_t entry = bar_emul->passthrough;
-            vmm_pci_device_def_t *pci_config = (vmm_pci_device_def_t *)entry.cookie;
+        vmm_pci_entry_t *dev = pci->bus0[i][0];
+        if (dev) {
             struct pci_interrupt_map irq_map;
+            uint32_t interrupt_pin = 0;
+            uint32_t interrupt_line = 0;
             irq_map.pci_mask.pci_addr.hi  = cpu_to_fdt32(i << PCI_ADDR_DEV_SHIFT);
             irq_map.pci_mask.pci_addr.mid  = 0;
             irq_map.pci_mask.pci_addr.low  = 0;
-            irq_map.pci_mask.irq_pin = cpu_to_fdt32(pci_config->interrupt_pin);
+            if (dev->ioread(dev->cookie, PCI_INTERRUPT_PIN, 1, &interrupt_pin)) {
+                ZF_LOGE("Error reading interrupt pin from PCI device");
+            }
+            if (dev->ioread(dev->cookie, PCI_INTERRUPT_LINE, 1, &interrupt_line)) {
+                ZF_LOGE("Error reading interrupt line from PCI device");
+            }
+            irq_map.pci_mask.irq_pin = cpu_to_fdt32(interrupt_pin);
             irq_map.gic_phandle = cpu_to_fdt32(gic_phandle);
             irq_map.irq_type = 0;
 #if GIC_ADDRESS_CELLS == 0x1
-            irq_map.irq_num = cpu_to_fdt32(pci_config->interrupt_line - 32);
+            irq_map.irq_num = cpu_to_fdt32(interrupt_line - 32);
 #else
-            irq_map.irq_num = cpu_to_fdt64(pci_config->interrupt_line - 32);
+            irq_map.irq_num = cpu_to_fdt64(interrupt_line - 32);
 #endif
             irq_map.irq_flags = cpu_to_fdt32(0x4);
             FDT_OP(fdt_appendprop(fdt, pci_node, "interrupt-map", &irq_map, sizeof(irq_map)));
--- a/libsel4vmmplatsupport/src/arch/arm/devices/vpci.c
+++ b/libsel4vmmplatsupport/src/arch/arm/devices/vpci.c
@@ -71,13 +71,12 @@ struct pci_cfg_data {
     vmm_pci_space_t *pci;
 };
 
-static void pci_cfg_read_fault(struct device *d, vm_t *vm, vm_vcpu_t *vcpu, vmm_pci_address_t pci_addr,
-                               uint8_t offset, vmm_pci_entry_t *dev)
+static void pci_cfg_read_fault(vm_vcpu_t *vcpu, uint8_t offset, size_t len, vmm_pci_entry_t *dev)
 {
     uint32_t data = 0;
     int err = 0;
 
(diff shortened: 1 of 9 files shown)


hlyytine:
See also https://github.com/seL4/camkes-vm/pull/18

hlyytine:
We are working on having the virtio backends on a separate guest with QEMU à la KVM. To accomplish this, we found out the PCI passthrough emulation does not work as expected on ARM. Fixed that and as a side effect eliminated some delta between x86 and ARM codebase.

libsel4vmmplatsupport/src/arch/arm/devices/vpci.c
```diff
@@ -283,21 +271,27 @@ int fdt_generate_vpci_node(vm_t *vm, vmm_pci_space_t *pci, void *fdt, int gic_ph
...
+            }
+            irq_map.pci_mask.irq_pin = cpu_to_fdt32(interrupt_pin);
             irq_map.gic_phandle = cpu_to_fdt32(gic_phandle);
             irq_map.irq_type = 0;
 #if GIC_ADDRESS_CELLS == 0x1
-            irq_map.irq_num = cpu_to_fdt32(pci_config->interrupt_line - 32);
+            irq_map.irq_num = cpu_to_fdt32(interrupt_line - 32);
 #else
-            irq_map.irq_num = cpu_to_fdt64(pci_config->interrupt_line - 32);
+            irq_map.irq_num = cpu_to_fdt64(interrupt_line - 32);
```
kent-mcleod, line 294: These changes look like a valid improvement.  However, can you please move them to a separate commit as they appear to be addressing a separate issue that isn't described in the commit message: That the code can't assume that `pci->bus0[i][0])->cookie` is always `pci_bar_emulation_t`

libsel4vmmplatsupport/src/drivers/pci_helper.c
```diff
@@ -255,19 +242,6 @@ vmm_pci_entry_t vmm_pci_create_bar_emulation(vmm_pci_entry_t existing, int num_b
...
-    assert(bar_emul);
-    memcpy(bar_emul->bars, bars, sizeof(vmm_pci_bar_t) * num_bars);
-    bar_emul->passthrough = existing;
-    bar_emul->num_bars = num_bars;
-    memset(bar_emul->bar_writes, 0, sizeof(bar_emul->bar_writes));
-    return (vmm_pci_entry_t) {
-        .cookie = bar_emul, .ioread = pci_bar_passthrough_emul_read, .iowrite = pci_bar_passthrough_emul_write
-    };
-}
-
```
kent-mcleod, line 270: Can you please move the removal of this function, it's implementation and it's documentation to a separate commit. Also in the commit message can you explain why it should be removed. (It appears to me that it's a no-op as it could be implemented by just returning `existing`).

libsel4vmmplatsupport/src/drivers/pci_helper.c
```diff
@@ -23,12 +23,11 @@ int vmm_pci_mem_device_read(void *cookie, int offset, int size, uint32_t *result
         ZF_LOGE("Offset should not be negative");
         return -1;
     }
-    if (offset + size >= PCI_CAPABILITY_SPACE_OFFSET) {
+    if (offset + size > PCI_CAPABILITY_SPACE_OFFSET) {
         ZF_LOGI("Indexing capability space not yet supported, returning 0");
         *result = 0;
         return 0;
     }
-    *result = 0;
```
kent-mcleod, line 31: Is this intentionally removed? I assume it was there to handle when `size < 4` and `*result` isn't 0.

kent-mcleod:
These are nice improvements! I just had some minor requests for organizing the changes into separate commits for each independent improvement.

> To accomplish this, we found out the PCI passthrough emulation does not work as expected on ARM.

Was the issue because the BAR emulation on Arm only handled BAR size and not also memory types or addressing?

libsel4vmmplatsupport/src/arch/arm/devices/vpci.c
hlyytine: Done, used uint32_t for those two variables, since the safety "set to zero" inside the ioread callback uses 32-bit write even if you wanted to read only one byte.

libsel4vmmplatsupport/src/drivers/pci_helper.c
hlyytine: Done.

libsel4vmmplatsupport/src/drivers/pci_helper.c
hlyytine: Yes it was. And yes, most probably you are correct. And that's a bit problematic. On the other hand the function prototype says the data pointer is uint32_t, but on the another, it also takes the length argument. An ignorant one like me could imagine it's OK to pass a pointer to uint8_t if you pass length = 1, right? But inside this function you get 32-bit write anyway.

I think the best solution would be to change the callback and function prototypes to use void * pointer. That way you could not make any assumptions about the read/write sizes, but you would need to explicitly rely on the length argument. But for now, I just changed the FDT generation code to use uint32_t and also assign the local variables there to zero beforehand, just in case this set-to-zero gets eventually deleted for some reason. So far this works because ARM is configured for little-endian, but that void pointer would make this endian-agnostic. Or so I believe.

TL;DR: added this back, worked around the issues in the FDT code.

hlyytine:
> Was the issue because the BAR emulation on Arm only handled BAR size and not also memory types or addressing?

So in our use case QEMU somewhere else emulates the whole PCI device, including this 64-byte configuration header and the BAR registers within it. The passthrough PCI emulation was ideal for our purposes, except that we found out someone was messing with the BAR registers in between the QEMU and the VM. So it turned out to be this BAR emulation implemented in the fault handler itself, leaving the PCI passthru unable to really pass through those registers unmodified (like it does on x86).

libsel4vmmplatsupport/src/drivers/pci_helper.c
hlyytine: I think you can find the explanation in the commit itself, just for sake of convenience that kind of emulation existed only to prevent doing double emulation on BAR registers; the fault handler did that anyway, so we had to avoid doing that in the emulation code before also.

kent-mcleod:
So when you are using passthrough from Qemu you provide your own `vmm_pci_entry_t *` object that isn't created by `vmm_pci_create_passthrough_bar_emulation` or `vmm_pci_create_bar_emulation`?  But because the vpci fault handler on Arm would always BAR emulation it was modifying the write data before it reached your passthrough handlers.  By removing this part of the Arm vpci fault handler it resolved your issue while also allowing enabling some simplification in the virtio device setup and the removal of vmm_pci_create_passthrough_bar_emulation to be removed?

Or is it that you wished to use `vmm_pci_create_bar_emulation` which is mostly passthrough, but still emulates the BAR address registers to allow remapping of IOPort and Memory regions?

libsel4vmmplatsupport/src/drivers/pci_helper.c
kent-mcleod: I agree with your comment. At the moment it's implemented in a way that tries to model the memory region as a 32bit CPU register. To allow 64bit reads/writes this would need to be modified.

kent-mcleod (approved):

hlyytine:
The former -- we are using the object created by `vmm_pci_create_passthrough`.

PR #28, opened 2022-03-02
libsel4vm: Enqueue IRQs only once

hlyytine:
The real GIC has exactly zero or one pending request for any IRQ.

Signed-off-by: Hannu Lyytinen <hannux@ssrc.tii.ae>

### DIFF
--- a/libsel4vm/src/arch/arm/vgic/vgic.c
+++ b/libsel4vm/src/arch/arm/vgic/vgic.c
@@ -611,6 +611,10 @@ static int vgic_dist_set_pending_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu
     virq_data = virq_find_irq_data(vgic, vcpu, irq);
     /* If it is enabled, inject the IRQ */
     if (virq_data && gic_dist->enable && is_enabled(gic_dist, irq, vcpu->vcpu_id)) {
+        if (is_pending(gic_dist, virq_data->virq, vcpu->vcpu_id)) {
+            return 0;
+        }
+
         int err;
         DDIST("Pending set: Inject IRQ from pending set (%d)\n", irq);
 
--- a/libsel4vm/src/arch/arm/vgic/vgic.c
+++ b/libsel4vm/src/arch/arm/vgic/vgic.c
@@ -70,7 +70,6 @@
 #define DDIST(...) do{}while(0)
 #endif
 
-
 /* FIXME these should be defined in a way that is friendlier to extension. */
 #if defined(CONFIG_PLAT_EXYNOS5)
 #define GIC_PADDR   0x10480000
@@ -217,8 +216,7 @@ static struct virq_handle *virq_get_sgi_ppi(vgic_t *vgic, vm_vcpu_t *vcpu, int v
 
 static struct virq_handle *virq_find_spi_irq_data(struct vgic *vgic, int virq)
 {
-    int i;
-    for (i = 0; i < MAX_VIRQS; i++) {
+    for (int i = 0; i < MAX_VIRQS; i++) {
         if (vgic->virqs[i] && vgic->virqs[i]->virq == virq) {
             return vgic->virqs[i];
         }
@@ -236,8 +234,7 @@ static struct virq_handle *virq_find_irq_data(struct vgic *vgic, vm_vcpu_t *vcpu
 
 static int virq_spi_add(vgic_t *vgic, struct virq_handle *virq_data)
 {
-    int i;
-    for (i = 0; i < MAX_VIRQS; i++) {
+    for (int i = 0; i < MAX_VIRQS; i++) {
         if (vgic->virqs[i] == NULL) {
             vgic->virqs[i] = virq_data;
             return 0;
@@ -265,7 +262,7 @@ static int virq_add(vm_vcpu_t *vcpu, vgic_t *vgic, struct virq_handle *virq_data
     return virq_spi_add(vgic, virq_data);
 }
 
-static int vgic_virq_init(vgic_t *vgic)
+static void vgic_virq_init(vgic_t *vgic)
 {
     memset(vgic->virqs, 0, sizeof(vgic->virqs));
     for (int i = 0; i < CONFIG_MAX_NUM_NODES; i++) {
@@ -274,9 +271,7 @@ static int vgic_virq_init(vgic_t *vgic)
         vgic->irq_queue[i].tail = 0;
         memset(vgic->irq_queue[i].irqs, 0, sizeof(vgic->irq_queue[i].irqs));
     }
-    return 0;
 }
-
 
 static inline void virq_init(virq_handle_t virq, int irq, irq_ack_fn_t ack_fn, void *token)
 {
@@ -284,7 +279,6 @@ static inline void virq_init(virq_handle_t virq, int irq, irq_ack_fn_t ack_fn, v
     virq->token = token;
     virq->ack = ack_fn;
 }
-
 
 static inline struct vgic *vgic_device_get_vgic(struct vgic_dist_device *d)
 {
@@ -331,17 +325,17 @@ static inline void set_pending(struct gic_dist_map *gic_dist, int irq, int value
     set_spi_pending(gic_dist, irq, value, vcpu_id);
 }
 
-static inline int is_sgi_ppi_pending(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_sgi_ppi_pending(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     return !!(gic_dist->pending_set0[vcpu_id] & IRQ_BIT(irq));
 }
 
-static inline int is_spi_pending(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_spi_pending(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     return !!(gic_dist->pending_set[IRQ_IDX(irq)] & IRQ_BIT(irq));
 }
 
-static inline int is_pending(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_pending(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     if (irq < GIC_SPI_IRQ_MIN) {
         return is_sgi_ppi_pending(gic_dist, irq, vcpu_id);
@@ -360,7 +354,6 @@ static inline void set_sgi_ppi_enable(struct gic_dist_map *gic_dist, int irq, in
         gic_dist->enable_clr0[vcpu_id] &= ~IRQ_BIT(irq);
     }
 }
-
 
 static inline void set_spi_enable(struct gic_dist_map *gic_dist, int irq, int value, int vcpu_id)
 {
@@ -382,17 +375,17 @@ static inline void set_enable(struct gic_dist_map *gic_dist, int irq, int value,
     set_spi_enable(gic_dist, irq, value, vcpu_id);
 }
 
-static inline int is_sgi_ppi_enabled(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_sgi_ppi_enabled(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     return !!(gic_dist->enable_set0[vcpu_id] & IRQ_BIT(irq));
 }
 
-static inline int is_spi_enabled(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_spi_enabled(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     return !!(gic_dist->enable_set[IRQ_IDX(irq)] & IRQ_BIT(irq));
 }
 
-static inline int is_enabled(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_enabled(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     if (irq < GIC_SPI_IRQ_MIN) {
         return is_sgi_ppi_enabled(gic_dist, irq, vcpu_id);
@@ -427,17 +420,17 @@ static inline void set_active(struct gic_dist_map *gic_dist, int irq, int value,
     set_spi_active(gic_dist, irq, value, vcpu_id);
 }
 
-static inline int is_sgi_ppi_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_sgi_ppi_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     return !!(gic_dist->active0[vcpu_id] & IRQ_BIT(irq));
 }
 
-static inline int is_spi_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_spi_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     return !!(gic_dist->active[IRQ_IDX(irq)] & IRQ_BIT(irq));
 }
 
-static inline int is_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
+static inline bool is_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
 {
     if (irq < GIC_SPI_IRQ_MIN) {
         return is_sgi_ppi_active(gic_dist, irq, vcpu_id);
@@ -538,14 +531,11 @@ static int vgic_dist_disable(struct vgic_dist_device *d, vm_t *vm)
 
 static int vgic_dist_enable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, int irq)
 {
-    struct gic_dist_map *gic_dist;
-    struct virq_handle *virq_data;
-    vgic_t *vgic;
-    gic_dist = vgic_priv_get_dist(d);
-    vgic = vgic_device_get_vgic(d);
+    struct gic_dist_map *gic_dist = vgic_priv_get_dist(d);
+    vgic_t *vgic = vgic_device_get_vgic(d);
     DDIST("enabling irq %d\n", irq);
     set_enable(gic_dist, irq, true, vcpu->vcpu_id);
-    virq_data = virq_find_irq_data(vgic, vcpu, irq);
+    struct virq_handle *virq_data = virq_find_irq_data(vgic, vcpu, irq);
     if (virq_data) {
         /* STATE b) */
         if (!is_pending(gic_dist, virq_data->virq, vcpu->vcpu_id)) {
@@ -1017,28 +1007,26 @@ static void vgic_dist_reset(struct vgic_dist_device *d)
     gic_dist->component_id[1] = 0x000000f0; /* RO */
     gic_dist->component_id[2] = 0x00000005; /* RO */
     gic_dist->component_id[3] = 0x000000b1; /* RO */
-
 }
 
 int vm_register_irq(vm_vcpu_t *vcpu, int irq, irq_ack_fn_t ack_fn, void *cookie)
 {
-    struct virq_handle *virq_data;
-    struct vgic *vgic;
-    int err;
-
-    vgic = vgic_device_get_vgic(vgic_dist);
+    struct vgic *vgic = vgic_device_get_vgic(vgic_dist);
     assert(vgic);
 
-    virq_data = calloc(1, sizeof(*virq_data));
+    struct virq_handle *virq_data = calloc(1, sizeof(*virq_data));
     if (!virq_data) {
         return -1;
     }
+
     virq_init(virq_data, irq, ack_fn, cookie);
-    err = virq_add(vcpu, vgic, virq_data);
+
+    int err = virq_add(vcpu, vgic, virq_data);
     if (err) {
         free(virq_data);
         return -1;
     }
+
     return 0;
 }
 
@@ -1070,12 +1058,11 @@ static memory_fault_result_t handle_vgic_vcpu_fault(vm_t *vm, vm_vcpu_t *vcpu, u
 
 static vm_frame_t vgic_vcpu_iterator(uintptr_t addr, void *cookie)
 {
-    int err;
     cspacepath_t frame;
     vm_frame_t frame_result = { seL4_CapNull, seL4_NoRights, 0, 0 };
     vm_t *vm = (vm_t *)cookie;
 
-    err = vka_cspace_alloc_path(vm->vka, &frame);
+    int err = vka_cspace_alloc_path(vm->vka, &frame);
     if (err) {
         printf("Failed to allocate cslot for vgic vcpu\n");
         return frame_result;
@@ -1102,20 +1089,12 @@ static vm_frame_t vgic_vcpu_iterator(uintptr_t addr, void *cookie)
  */
 int vm_install_vgic(vm_t *vm)
 {
-    struct vgic *vgic;
-    void *addr;
-    int err;
-
-    vgic = calloc(1, sizeof(*vgic));
+    struct vgic *vgic = calloc(1, sizeof(*vgic));
     if (!vgic) {
         assert(!"Unable to calloc memory for VGIC");
         return -1;
     }
-    err = vgic_virq_init(vgic);
-    if (err) {
-        free(vgic);
-        return -1;
-    }
+    vgic_virq_init(vgic);
 
     /* Distributor */
     vgic_dist = (struct vgic_dist_device *)calloc(1, sizeof(struct vgic_dist_device));
@@ -1137,7 +1116,7 @@ int vm_install_vgic(vm_t *vm)
     /* Remap VCPU to CPU */
     vm_memory_reservation_t *vgic_vcpu_reservation = vm_reserve_memory_at(vm, GIC_CPU_PADDR,
                                                                           0x1000, handle_vgic_vcpu_fault, NULL);
-    err = vm_map_reservation(vm, vgic_vcpu_reservation, vgic_vcpu_iterator, (void *)vm);
+    int err = vm_map_reservation(vm, vgic_vcpu_reservation, vgic_vcpu_iterator, (void *)vm);
     if (err) {
         free(vgic_dist->priv);
         return -1;
@@ -1148,13 +1127,11 @@ int vm_install_vgic(vm_t *vm)
 
 int vm_vgic_maintenance_handler(vm_vcpu_t *vcpu)
 {
-    int idx;
-    int err;
-    idx = seL4_GetMR(seL4_UnknownSyscall_ARG0);
+    int idx = seL4_GetMR(seL4_UnknownSyscall_ARG0);
     /* Currently not handling spurious IRQs */
     assert(idx >= 0);
 
-    err = handle_vgic_maintenance(vcpu, idx);
+    int err = handle_vgic_maintenance(vcpu, idx);
     if (!err) {
         seL4_MessageInfo_t reply;
         reply = seL4_MessageInfo_new(0, 0, 0, 0);
--- a/libsel4vm/src/arch/arm/vgic/vgic.c
+++ b/libsel4vm/src/arch/arm/vgic/vgic.c
@@ -110,10 +110,6 @@
 #define IRQ_IDX(irq) ((irq) / 32)
 #define IRQ_BIT(irq) (1U << ((irq) % 32))
 
-#define not_pending(...) !is_pending(__VA_ARGS__)
-#define not_active(...)  !is_active(__VA_ARGS__)
-#define not_enabled(...) !is_enabled(__VA_ARGS__)
-
 struct virq_handle {
     int virq;
     irq_ack_fn_t ack;
@@ -121,9 +117,6 @@ struct virq_handle {
 };
 
 typedef struct virq_handle *virq_handle_t;
-
-/* Inject interrupt into vcpu */
-static int vgic_vcpu_inject_irq(struct vgic_dist_device *d, vm_vcpu_t *inject_vcpu, struct virq_handle *irq);
 
 static inline void virq_ack(vm_vcpu_t *vcpu, struct virq_handle *irq)
 {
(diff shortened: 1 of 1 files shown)


hlyytine:
It appears that vgic_vcpu_inject_irq() always succeeded, either by succeeding to load the IRQ into hardware-backed LR registers (4) or if that fails, into the software-extended 64 LR registers. Because it always succeeded, vgic_handle_overflow_cpu() (part of vGIC maintenance code) would always consume the queued IRQ. But in case the vgic_vcpu_inject_irq(), the consumed IRQ was appended back to software LR registers.

The real problem was that there was not any check whether or not the IRQ was already loaded into of these list registers. I have had a fix for this already, but it was rather clumsy. Now I realized a simple refactoring is all that is needed.

A little background: we have been playing with Raspberry Pi 4 for some time now and found out virtio-console makes this code bail out, saying the software-extended LRs are overflowing. The more cores you enable in the guest VM, the sooner that happens (with single core it happens never in practice, with four core it is a matter of seconds). This fix is a big improvement to that.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -591,6 +589,19 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
     return 0;
 }
 
+static int vgic_irq_queued(vgic_t *vgic, vm_vcpu_t *vcpu, int irq)
```
axel-h, line 592: nitpicking: There is `bool`, so this could start using it:
```suggestion
static bool vgic_is_irq_queued(vgic_t *vgic, vm_vcpu_t *vcpu, int irq)
```
and then return `true` or `false`

axel-h, line 592: nitpicking: There is `bool`, so this could start using it:
```suggestion
static bool vgic_is_irq_queued(vgic_t *vgic, vm_vcpu_t *vcpu, int irq)
```
and then return `true` or `false`

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -604,14 +615,17 @@ static int vgic_dist_set_pending_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu
...
-        assert(!err);
-
-        return err;
+        if (vgic_irq_queued(vgic, vcpu, irq)) {
+            return 0;
+        }
+        if (vgic_vcpu_inject_irq(d, vcpu, virq_data)) {
+            /* Add to overflow list */
+            int err = vgic_add_overflow(vgic, virq_data, vcpu);
+            assert(!err);
```
axel-h, line 627: Please add a comment here that it's really intended to swallow the error in release builds and no report anything to the caller.

axel-h, line 627: Please add a comment here that it's really intended to swallow the error in release builds and no report anything to the caller.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -604,14 +615,17 @@ static int vgic_dist_set_pending_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu
...
 
         set_pending(gic_dist, virq_data->virq, true, vcpu->vcpu_id);
-        err = vgic_vcpu_inject_irq(d, vcpu, virq_data);
-        assert(!err);
-
-        return err;
+        if (vgic_irq_queued(vgic, vcpu, irq)) {
+            return 0;
+        }
+        if (vgic_vcpu_inject_irq(d, vcpu, virq_data)) {
```
axel-h, line 624: I consider statements like this very easy to misinterpret, because it's doing the opposite actually.
```suggestion
        int err = vgic_vcpu_inject_irq(d, vcpu, virq_data);
        if (err) { ...
        
```

axel-h, line 624: I consider statements like this very easy to misinterpret, because it's doing the opposite actually.
```suggestion
        int err = vgic_vcpu_inject_irq(d, vcpu, virq_data);
        if (err) { ...
        
```

axel-h:
> It appears that vgic_vcpu_inject_irq() always succeeded, either by succeeding to load the IRQ into hardware-backed LR registers (4) or if that fails, into the software-extended 64 LR registers

Not sure I follow, but the list could be full (`unlikely(lr_overflow->full)`), so this could still fail in theory in extreme situations unless there is a guarantee the list can really holds all possible (1021?) interrupts.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -512,11 +512,9 @@ static int vgic_vcpu_inject_irq(struct vgic_dist_device *d, vm_vcpu_t *inject_vc
     if (!err) {
         /* Shadow */
         vgic->irq[inject_vcpu->vcpu_id][i] = irq;
-        return err;
-    } else {
-        /* Add to overflow list */
-        return vgic_add_overflow(vgic, irq, inject_vcpu);
     }
+
```
axel-h, line 516: Putting an `assert(!vgic_irq_queued(vgic, vcpu, irq))` might be too expensive, but this function should get a comment that that at this stage the callmer must have ensured the IRQ is not queued and thus there is no need to clean the queue here. I have been working on this code also and more comment everywhere are really helpful.

axel-h:
From my side this seems the proper fix and I'd consider this as a bug.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Not nitpicking at all IMO, but I realized there's is_pending() that we can use. Remove this addition altogether.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Actually I think that's not what we want. Failing to add the IRQ to overflow list is a critical and usually non-recoverable condition, so I modified the code to emit the warning message and pass the error in both builds.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Good idea. Did that here and the other places as well.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: In this case, the comment you suggested would not be entirely correct. Sure, if it is called from vgic_dist_set_pending_irq(), it is true, but when called by the vGIC maintenance interrupt, the IRQ is in the overflow queue; otherwise we wouldn't be trying to reinject that in the first place. Instead, I rewrote code and tried to made it clear that we are only trying to load the IRQ into one of the four hardware-backed list registers and that's all.

hlyytine:
> > It appears that vgic_vcpu_inject_irq() always succeeded, either by succeeding to load the IRQ into hardware-backed LR registers (4) or if that fails, into the software-extended 64 LR registers
> 
> Not sure I follow, but the list could be full (`unlikely(lr_overflow->full)`), so this could still fail in theory in extreme situations unless there is a guarantee the list can really holds all possible (1021?) interrupts.

It is entirely possible in theory, but I would refrain addressing that issue now. If we continue using the ring buffer for overflowed IRQs, it should be just a matter of enlarging the buffer. But perhaps some day we would like to emulate more recent GICs also and in the future there are more and more ARM chips that support up to 8192 interrupt IDs.

Another thing is whether or not we want to support IRQ priorities some day. For that the ring buffer is not the optimal data structure IMO and for example, priority queue based on binary heap could be more proper solution. That's why I would like to leave it as it is now and guest Linux does not need priorities anyway.

What I meant to say in the text you commented was that the ill-named vgic_vcpu_inject_irq() does not fail normally, it either succeeds loading the IRQ into HW registers (or to the abstraction for them within seL4 kernel but that's another story) or if that fails, it succeeds appending the IRQ to overflow list. And this "always succeeds" resulted in unnecessary shuffling of IRQs out of that list, just to be put into that again.

The first iteration that I subjected to review had to solve that problem in order to figure out whether or not the IRQ was queued. But now I realized we could use is_pending() instead and the whole bugfix would reduce to a one-liner. But since it appears we agree that the code is rather complex, I took some liberties and renamed variables, fields and functions and even added one commit which I believe adds even more readability. I'm happy to add more comments but I believe the code itself must be improved first.

hlyytine:
Any opinions whether I should split the first commit into two, one having only this "if (is_pending(...))" (the real fix) and another having the renames. The third that gets rid of per-CPU helper functions is perhaps best to left unmodified?

Ivan-Velickovic:
The number of list registers (`NUM_LRS`) is implementation defined and so isn't the same across all systems (although 4 seems to be the common number), would it be worth adding a comment above the `#define` noting this?

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -122,8 +122,8 @@ struct virq_handle {
 
 typedef struct virq_handle *virq_handle_t;
 
-/* Inject interrupt into vcpu */
-static int vgic_vcpu_inject_irq(struct vgic_dist_device *d, vm_vcpu_t *inject_vcpu, struct virq_handle *irq);
+/* Load IRQ into hardware list register */
+static int vgic_vcpu_load_lr(struct vgic_dist_device *d, vm_vcpu_t *vcpu, struct virq_handle *irq);
```
axel-h, line 126: I thing we don't need this prototype here, the function is only called after the implementation.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -182,6 +182,7 @@ struct gic_dist_map {
     uint32_t component_id[4];                           /* [0xFF0, 0xFFF] */
 };
 
+#define NUM_LRS 4
```
axel-h, line 185: To anser https://github.com/seL4/seL4_projects_libs/pull/28#issuecomment-1063571957, yes, would be nice to have a comment here where the value comes from.

libsel4vm/src/arch/arm/vgic/vgic.c
axel-h: ah no, we need it

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -467,56 +469,42 @@ static inline int vgic_add_overflow_cpu(struct lr_of *lr_overflow, struct virq_h
     return 0;
 }
 
-static inline int vgic_add_overflow(vgic_t *vgic, struct virq_handle *irq, vm_vcpu_t *vcpu)
+static inline void vgic_handle_overflow(vgic_t *vgic, vm_vcpu_t *vcpu)
```
axel-h, line 472: Could call this `vgic_process_lr_queue` now, because this is what is is doing effectively

axel-h: Actually, why not fully inline this completely, the maintenance routine is the only place that calls this

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -467,56 +469,42 @@ static inline int vgic_add_overflow_cpu(struct lr_of *lr_overflow, struct virq_h
...
-            lr_overflow->head = LR_OF_NEXT(i);
-            lr_overflow->full = (lr_overflow->head == LR_OF_NEXT(lr_overflow->tail));
-        } else {
+    for (size_t i = lr_overflow->head; i != lr_overflow->tail; i = LR_OF_NEXT(i)) {
+        int err = vgic_vcpu_load_lr(vgic_dist, vcpu, &lr_overflow->irqs[i]);
+        if (err) {
             break;
         }
+        lr_overflow->head = LR_OF_NEXT(i);
+        lr_overflow->full = (lr_overflow->head == LR_OF_NEXT(lr_overflow->tail));
```
axel-h, line 482: Can we simply say
```suggestion
        lr_overflow->full = false; /* at least 1 slot is free now */
```
here, because when we removed an element from the queue, this implies the queue is no longer full.

libsel4vm/src/arch/arm/vgic/vgic.c
axel-h: or well, just change the order in the cleanup and we can drop this prototype and get rid of some redundancy in the code

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -594,27 +582,32 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
...
-        return err;
-    } else {
-        /* No further action */
-        DDIST("IRQ not enabled (%d) on vcpu %d\n", irq, vcpu->vcpu_id);
+    DDIST("Pending set: Inject IRQ from pending set (%d)\n", irq);
+    set_pending(gic_dist, virq_data->virq, true, vcpu->vcpu_id);
+
+    int err = vgic_vcpu_load_lr(d, vcpu, virq_data);
+    if (!err) {
+        return 0;
```
axel-h, line 604: The return code of `vgic_vcpu_load_lr()` is basically the one from `seL4_ARM_VCPU_InjectIRQ()`:
- seL4_DeleteFirst: this seems save to ignore, add to lr overflow queue
- seL4_IllegalOperation:should bubble up as a fatal error?
- seL4_InvalidCapability: this should bubble up as a fatal error?
- seL4_RangeError: this should bubble up as a fatal error?

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -594,27 +582,32 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
...
+    }
+
+    /* Add to overflow list */
+    err = vgic_add_overflow(vgic, virq_data, vcpu);
```
axel-h, line 608: We could inline the function here, this is the only place where this is called.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -194,7 +195,7 @@ struct lr_of {
 
 typedef struct vgic {
 /// Mirrors the vcpu list registers
-    struct virq_handle *irq[CONFIG_MAX_NUM_NODES][MAX_LR_OVERFLOW - 1];
+    struct virq_handle *lrs[CONFIG_MAX_NUM_NODES][NUM_LRS];
```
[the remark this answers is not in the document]
axel-h: ah no, this is `lr_shadow`

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Done.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Done.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Please check my new approach; personally I don't like inlining, I want to hide the implementation details especially in those are completely orthogonal to the workings of the caller.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Yes.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: With the latest refactoring, all errors from seL4 kernel are considered fatal.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: See my comment above about inlining.

hlyytine:
Redid the patch, personally I think it's one more step towards readability, let me know what you think.

libsel4vm/src/arch/arm/vgic/vgic.c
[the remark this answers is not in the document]
hlyytine: Just named it `link_regs`, OK?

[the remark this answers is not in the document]
hlyytine: Actually reconsidered this and found this good idea. Changed also variable name in the vGIC maintenance handler to reflect the fact it is only a shadow of the real LR.

kent-mcleod:
> Any opinions whether I should split the first commit into two, one having only this "if (is_pending(...))" (the real fix) and another having the renames? The third that gets rid of per-CPU helper functions is perhaps best to left unmodified?

It would make it easier to review and integrate these changes if they were in separate commits.  (There's something like 4 or 5 different branches that implement gicv3 virt that may need to cherry-pick or rebase over these changes).

hlyytine:
> > Any opinions whether I should split the first commit into two, one having only this "if (is_pending(...))" (the real fix) and another having the renames? The third that gets rid of per-CPU helper functions is perhaps best to left unmodified?
> 
> It would make it easier to review and integrate these changes if they were in separate commits. (There's something like 4 or 5 different branches that implement gicv3 virt that may need to cherry-pick or rebase over these changes).

That's my thinking as well. Right now I squashed all my previous changes into one, because the intermediate commits made some changes just to be changed again in later commits, so I hope this squashed commit is easier to review. After we agree on the contents, I can proceed with splitting the PR into several commits.

hlyytine:
Removed unused macros, simplified IRQ queue ring buffer, explained a bit better how the number of list registers could be autoprobed.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -450,99 +445,83 @@ static inline int is_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
     return is_spi_active(gic_dist, irq, vcpu_id);
 }
 
-static inline int vgic_add_overflow_cpu(struct lr_of *lr_overflow, struct virq_handle *irq)
+static inline int vgic_irq_enqueue(vgic_t *vgic, vm_vcpu_t *vcpu, struct virq_handle *irq)
```
axel-h, line 448: Let's have this return a bool then also, as we don't have any error codes.
```suggestion
static inline boot vgic_irq_enqueue(vgic_t *vgic, vm_vcpu_t *vcpu, struct virq_handle *irq)
```

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -450,99 +445,83 @@ static inline int is_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
...
     int i;
 
-    vgic = vgic_device_get_vgic(d);
-
-    seL4_CPtr vcpu;
-    vcpu = inject_vcpu->vcpu.cptr;
-    for (i = 0; i < MAX_LR_OVERFLOW; i++) {
-        if (vgic->irq[inject_vcpu->vcpu_id][i] == NULL) {
-            break;
+    for (i = 0; i < NUM_LIST_REGS; i++) {
```
axel-h, line 479: See we only need the loop variable inthe body now, so let's use c99 features:
```suggestion
    for (int i = 0; i < NUM_LIST_REGS; i++) {
```

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -450,99 +445,83 @@ static inline int is_active(struct gic_dist_map *gic_dist, int irq, int vcpu_id)
...
+
+    return -1;
+}
+
+static int vgic_vcpu_load_list_reg(vgic_t *vgic, vm_vcpu_t *vcpu, int idx, struct virq_handle *irq)
+{
+    int err = seL4_ARM_VCPU_InjectIRQ(vcpu->vcpu.cptr, irq->virq, 0, 0, idx);
+    if (err) {
+        ZF_LOGF("Failure loading vGIC list register");
+        return -1;
```
axel-h, line 493: return `err`here to give the caller a chance to act on the error details?

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -594,31 +573,41 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
...
+    DDIST("Pending set: Inject IRQ from pending set (%d)\n", irq);
+    set_pending(gic_dist, virq_data->virq, true, vcpu->vcpu_id);
+
+    /* Enqueueing an IRQ and dequeueing it right after makes little sense
+     * now, but in the future this is needed to support IRQ priorities.
+     */
+    int err = vgic_irq_enqueue(vgic, vcpu, virq_data);
+    if (err) {
+        ZF_LOGF("Too many queued IRQs");
         return -1;
```
axel-h, line 599: same here, return `err` here?

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -594,31 +573,41 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
...
-        DDIST("IRQ not enabled (%d) on vcpu %d\n", irq, vcpu->vcpu_id);
+    DDIST("Pending set: Inject IRQ from pending set (%d)\n", irq);
+    set_pending(gic_dist, virq_data->virq, true, vcpu->vcpu_id);
+
+    /* Enqueueing an IRQ and dequeueing it right after makes little sense
+     * now, but in the future this is needed to support IRQ priorities.
+     */
+    int err = vgic_irq_enqueue(vgic, vcpu, virq_data);
+    if (err) {
+        ZF_LOGF("Too many queued IRQs");
```
axel-h, line 598: Can't be bad to print the error also, so we have it in the logs:
```suggestion
        ZF_LOGF("Enquequing IRQ failed, code %d", err);
```

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -594,31 +573,41 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
...
         return -1;
     }
 
-    return 0;
+    int idx = vgic_find_empty_list_reg(vgic, vcpu);
+    if (idx < 0) {
+        return 0;
```
axel-h, line 604: Can you add a brief comment here, stating that this basically means we keep the IRQ in the overflow list, which will be drained eventually. Can be bad to have a bit of explanation what this going on.

axel-h:
Thanks for the work, the code is really getting nicer now.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Done, also changed int to bool in many other places also.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Did this on other places too.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Yes.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: We don't have anything specific to report -- changed the error message a bit more descriptive instead.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Now suggesting the user to increase ring buffer size.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: Done.

hlyytine:
Thank you for your suggestions, I applied them to whole file and added two commits -- one that addresses the locations you mentioned and the other one addresses the rest. Kept them separate now for clarity.

hlyytine:
Just pondering; shouldn't we reserve the `bool` return type to predicates (function name should contain is or some adjective)? Thinking about this `vgic_irq_enqueue()`, the function name or the if statement containing that is not asking about the truth value, therefore it feels a bit funny for it to answer true or false to something that in expressed in an imperative mood. Maybe old fashioned zero on success, non-zero otherwise would be better together with your idea of first assigning it to `err` variable and then checking it?

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -432,117 +420,99 @@ static inline void set_active(struct gic_dist_map *gic_dist, int irq, int value,
...
+    virq_ack(vcpu, lr_shadow[idx]);
+    lr_shadow[idx] = NULL;
 
     /* Check the overflow list for pending IRQs */
-    lr[idx] = NULL;
-    vgic_t *vgic = vgic_device_get_vgic(vgic_dist);
-    vgic_handle_overflow(vgic, vcpu);
+    struct virq_handle *virq = vgic_irq_dequeue(vgic, vcpu);
+    if (virq) {
+        return vgic_vcpu_load_list_reg(vgic, vcpu, idx, virq);
```
axel-h, line 510: I wonder, if we should print an error message here, because this would indicate a serious problem in the maintenance operation

```suggestion
        int err = vgic_vcpu_load_list_reg(vgic, vcpu, idx, virq);
        if (err) {
            ZF_LOGE("IRQ injection failed, code %d\n", err);
        }
        return err
```

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -1156,13 +1127,11 @@ int vm_install_vgic(vm_t *vm)
...
 {
-    int idx;
-    int err;
-    idx = seL4_GetMR(seL4_UnknownSyscall_ARG0);
+    int idx = seL4_GetMR(seL4_UnknownSyscall_ARG0);
     /* Currently not handling spurious IRQs */
     assert(idx >= 0);
 
-    err = handle_vgic_maintenance(vcpu, idx);
+    int err = handle_vgic_maintenance(vcpu, idx);
```
axel-h, line 1134: Maybe we should also log a message here if this fails, as things are badly broken then. See also comment https://github.com/seL4/seL4_projects_libs/pull/28/files#r824617661

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: But `vgic_vcpu_load_list_reg()` will print an error message regardless who calls it?

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: I think we should have a way to halt the VM execution at this point -- I think it should execute correctly or not execute at all, something in between only results in frustation and wasted time debugging something flaky.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -432,117 +420,99 @@ static inline void set_active(struct gic_dist_map *gic_dist, int irq, int value,
...
-        }
-    }
-    err = seL4_ARM_VCPU_InjectIRQ(vcpu, irq->virq, 0, 0, i);
-    assert((i < 4) || err);
-    if (!err) {
-        /* Shadow */
-        vgic->irq[inject_vcpu->vcpu_id][i] = irq;
+    int err = seL4_ARM_VCPU_InjectIRQ(vcpu->vcpu.cptr, irq->virq, 0, 0, idx);
+    if (err) {
+        ZF_LOGF("Failure loading vGIC list register");
```
axel-h, line 483:
```suggestion
        ZF_LOGF("Failure loading vGIC list register, code %d", err);
```

libsel4vm/src/arch/arm/vgic/vgic.c
axel-h: yes, you are right. It would make a nice error stack though seeing this bubble up though the layers then.

hlyytine: Now that is something I can agree.

libsel4vm/src/arch/arm/vgic/vgic.c
axel-h: It's a lost interrupt in this case, which is not necessarily this fatal, but very annoying. Logging is the best one can do to allow post-mortem analysis. The error behavior should become a follow-up discussion to leave out od this PR, otherwise this never gets merged ;)

hlyytine: True -- actually that reminds me I have pondered whether these readability improvements should be a separate PR (or moving the real bugfix into its own PR).

axel-h: From my side, having this in a separate commit is sufficient. It's more about how to get the others to agree to merge all the improvements, there are a few other PRs now also that seem useful (https://github.com/seL4/seL4_projects_libs/pull/29, https://github.com/seL4/seL4_projects_libs/pull/31 ...)
Opinions, @yyshen, @kent-mcleod, @lsf37

hlyytine:
Pushed one more commit that reverts `vgic_irq_enqueue()` using zero/non-zero return values ar per discussed above. Now I think I'm done, waiting for further discussion.

wom-bat:
@hlyytine is it your intention to split this into separate PRs or are you happy for this one to be updated and merged?

wom-bat (changes requested):
This looks good; but needs rebasing to make it mergeable.

hlyytine:
> @hlyytine is it your intention to split this into separate PRs or are you happy for this one to be updated and merged?

Yes, I reorganized these into three commits: one for bugfix, one for refactoring in general and last one for style issues (mainly converting to C99 as suggested by @axel-h).

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -182,28 +174,31 @@ struct gic_dist_map {
...
-struct lr_of {
-    struct virq_handle irqs[MAX_LR_OVERFLOW]; /* circular buffer */
+struct irq_queue {
+    struct virq_handle *irqs[MAX_IRQ_QUEUE_LEN]; /* circular buffer */
     size_t head;
     size_t tail;
-    bool full;
 };
 
-static inline size_t LR_OF_NEXT(size_t idx)
```
axel-h, line 194: Please leave this in unless there is a strong reason to go back to the macro.

hlyytine: I can do that, but I would like you to first review my patch set fully; I'm getting rid of that `full` field of `struct lr_of` and ring buffer code is much more comprehensible now IMHO. I was thinking my `static_assert `about the requirement of size of power of two and the and-operation for finding out the next index would be self-documenting. TBH I found your addition not exactly helping understanding the code and feel that the data structure is implemented in non-optimal way if it needs helpers/comments like that. But not really wanting to start a flame war, I just would you to read all my code first.

axel-h: Sorry, my bad. Looked like a merge conflict to me first, but with the 2^n size this is better. 
My argument for the function is, that this allows a strong binding to the actual array size in the struct and not just a constant. It's supposed to make the code more robust and the compiler is able to optimize this like the macro.

axel-h: But that stronger array binding could also be done in the macro. I may make a separate PR for this, so it's no blocker for this PR.

hlyytine: Yeah, my intention was to ping you and ask how do you like my proposal, but you were faster. And I can see where you are coming from. We could have sanity checks with this 2^n also, but I guess then it boils down to more fundamental questions like a) what else should we assert and b) shouldn't we be doing the assert before using the index (right before dereferencing an array), not just when derive a new index.

Ideally the data structures should be in a separate header/source file, having them in between the code that uses them makes it quite hard to grasp sometimes. But something like that is a PITA to do in C.

hlyytine: Forgot to mention that I am not that worried about performance in this context. It is just readability that matters to me.

libsel4vm/src/arch/arm/vgic/vgic.c
hlyytine: So based on our discussion about using the bool only for predicate-like function names, I changed this back to using int, zero for success and non-zero otherwise. Just for the sake of consistency.

kent-mcleod (approved):
I also think this is a nice improvement.  My only request is that the middle commit message has a description of the irq_queue related changes also mentions that this sets up for supporting interrupt priorities in the future as well as allowing the maintenance handling code to become simpler.

libsel4vm/src/arch/arm/vgic/vgic.c
kent-mcleod: > My argument for the function is, that this allows a strong binding to the actual array size in the struct and not just a constant. It's supposed to make the code more robust and the compiler is able to optimize this like the macro.

I think this issue could be resolved outside of this PR.  The macro and array definitions are defined within 10 lines of each other and use the same size constant which is already good enough.

hlyytine:
True, added that now.

PR #8, opened 2019-08-09
 GICv3 emulation

pingerino:
This PR preps the vgic code in libsel4arm-vmm for gicv3 support with minimal code duplication and introduces gicv3 emulation.

It also cleans up the code a lot, to make it easier to read and maintain and pulls in some fixes by 
@chrisguikema from the dornerworks branches.

PR #29, opened 2022-03-08
Various VGIC fixes/clean up

Ivan-Velickovic:
A couple fixes:

1. `active_clr` has 31 registers that aren't banked, rather than 32.
2. Making the VGIC list registers for each VCPU be `MAX_LR_OVERFLOW` instead of `MAX_LR_OVERFLOW - 1`. This was just done for consistency, the current code is still correct since we only use the first four indexes (all other IRQs get added to the LR overflow array).
3. Clean up error checking in `vm_install_vgic`. I've just made it consistent and only used asserts since this happens early in the process of starting a VM.


libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -151,7 +151,7 @@ struct gic_dist_map {
     uint32_t active0[CONFIG_MAX_NUM_NODES];             /* [0x300, 0x304) */
     uint32_t active[31];                                /* [0x300, 0x380) */
     uint32_t active_clr0[CONFIG_MAX_NUM_NODES];         /* [0x380, 0x384) */
-    uint32_t active_clr[32];                            /* [0x384, 0x400) */
+    uint32_t active_clr[31];                            /* [0x384, 0x400) */
```
axel-h, line 154: Seems this has always been a bug then and the comments about the offsets were never right from this place on? Are we just lucky the code every worked because nobody uses this correctly?

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -1115,28 +1115,17 @@ int vm_install_vgic(vm_t *vm)
     int err;
 
     vgic = calloc(1, sizeof(*vgic));
-    if (!vgic) {
-        assert(!"Unable to calloc memory for VGIC");
-        return -1;
-    }
+    assert(vgic);
```
axel-h, line 1118: Not sure is you can do that, there is no global guarantee `calloc()` will be successful and asserts are removed in release code. Now one could argue that it never fails in release code because this is well tested - but still, you don't know the context this is used in. So I'd prefer to keep this and have an error path here that the caller must deal with.
Another option would be making all this static, so the compiler/linker can ensure this. How much dynamism do we need there in the end?

libsel4vm/src/arch/arm/vgic/vgic.c
Ivan-Velickovic: It's not a bug since the only access to that field is in the distributor fault handling code, and those accesses are dependent on the offsets defined in `vgicv2_defs.h`, which are correct. I don't think it was ever possible to access the last element when it was length 32.

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -1146,10 +1135,7 @@ int vm_install_vgic(vm_t *vm)
     vm_memory_reservation_t *vgic_vcpu_reservation = vm_reserve_memory_at(vm, GIC_CPU_PADDR,
                                                                           0x1000, handle_vgic_vcpu_fault, NULL);
     err = vm_map_reservation(vm, vgic_vcpu_reservation, vgic_vcpu_iterator, (void *)vm);
-    if (err) {
-        free(vgic_dist->priv);
-        return -1;
-    }
+    assert(!err);
```
axel-h, line 1138: We should keep the cleanup and just accept that thing might fail. It's the caller problem then to handle this.

axel-h:
I'm not sure about removing the error checking - why remove this?

libsel4vm/src/arch/arm/vgic/vgic.c
axel-h: I thing this was a bug, because if we have 32 words here, this uses offset 0x384 - 0x404 actually, so everything after this is 4 byte off.

Ivan-Velickovic: Where are we depending on these offsets?

axel-h: Maybe I'm missing something here. Is this describing some actual hardware or is this is virtual peripheral where the actual memory layout does not matter because it only ever accesses via the filed names, so the compiler handles this properly and any memory layout could be used?

Ivan-Velickovic: This is an emulated device so those addresses commented correspond to the actual hardware, not this register map. Since the code depends on pre-defined offsets and not the memory layout, I don't think this is a bug.

Ivan-Velickovic: But to answer your question, yes any memory layout could be used as long as the `#defines` in `vgicv2_defs.h` are correct.

axel-h: Let's add a 
```
compile_time_assert(gic_dist_map_size, 0x1000 == sizeo(struct gic_dist_map));
```
after the definition so the compiler can ensure things are as expected

Ivan-Velickovic: Since the register map is dependent on the number of VCPUs since some of the registers are banked, the size isn't always a page. But I guess I could do a slightly more complicated comptime assert.

axel-h: I wonder if we should remove all the offset comments and have a comment in the beginning that clearly states this is a pure virtual peripheral that just tries to mimic a hardware distributor. Then `CONFIG_MAX_NUM_NODES` is finde also. Otherwise the offsets there are confusing and not guaranteed. I guess this all came from a singe core implementation where this made sense, but more multiple (virtual) cores it does not.

axel-h: If this is virtual, all the "resX" could also be dropped, because nobody needs them and they don't guarantee anything when `CONFIG_MAX_NUM_NODES` changes.

Ivan-Velickovic: I'm implementing GICv3 right now, and was planning on adding a comment explaining why the register map is dependent on `CONFIG_MAX_NUM_NODES` and so the offsets aren't always correct. I think that they still have value, it's easier for me to go to the spec based on an offset commented rather than the name of the field.

Ivan-Velickovic: I'll add a comment here explaining it anyways.

Ivan-Velickovic: > If this is virtual, all the "resX" could also be dropped

Sure, I'll do that too.

axel-h: > offsets [...]  still have value, it's easier for me to go to the spec based on an offset commented rather than the name of the field.

Well, I'd argue here, that the specs [1][2] also use names, and these might be easier to search for than the number.

[1] https://developer.arm.com/documentation/ddi0471/b/programmers-model/distributor-register-summary
[2] https://developer.arm.com/documentation/ddi0516/e/programmers-model/distributor-register-summary

axel-h:
> I'm implementing GICv3 right now

The PR https://github.com/seL4/seL4_projects_libs/pull/8 also works towards this, could you check if you can take over this one?

Ivan-Velickovic:
> The PR #8 also works towards this, could you check if you can take over this one?

I am using #8 as a reference but was planning on making a separate PR. Is that okay or should I be trying to resurrect #8?

axel-h:
> I am using #8 as a reference but was planning on making a separate PR. Is that okay or should I be trying to resurrect #8?

if you already had a look there and concluded there's not much in it, then we can also close it as outdated and reference this PR as the new approach to add GICv3 support eventually.

Ivan-Velickovic:
> if you already had a look there and concluded there's not much in it, then we can also close it as outdated and reference this PR as the new approach to add GICv3 support eventually.

It's not that there's not much in it, it's that due to the code changing over time since the PR, I would rather selectively take what I need rather than take over the existing one.

axel-h:
> But if you think it's better to take over it, I can do that.

I have not really opinion there. Cherry-picking what's useful sound like a good plan. And once there's a new PR from you we close #8 one as obsolete.

lsf37:
Yes, I'm also happy with that approach.

kent-mcleod:
I think this should be put on hold until after https://github.com/seL4/seL4_projects_libs/pull/28 is resolved as it looks to be refactoring the same code.

libsel4vm/src/arch/arm/vgic/vgic.c
kent-mcleod: I agree with @axel-h, the error checking should stay as this is library and needs to allow applications to make policy choices about whether to abort on errors or do something else.

Ivan-Velickovic: Yes, this was my mistake, I wrongly assumed that asserts were enabled in release mode. I'll make these changes once #28  goes through.

axel-h:
https://github.com/seL4/seL4_projects_libs/pull/28 is merged now, can you rebase this? Would be good to have the `trivial: fix number of `active_clr` registers` merged then also.

axel-h:
@Ivan-Velickovic: could you rebase this an resolve the conflicts, so we can merge it?

axel-h:
@Ivan-Velickovic: could you rebase this and resolve the conflicts, so we can merge it?

Ivan-Velickovic:
Done. Apologies for forgetting about this.

axel-h (approved):

PR #31, opened 2022-03-10
trivial: cleanup helper functions setting VGIC distributor registers

Ivan-Velickovic:
The SPI related helper functions in `vgic.c` don't need `vcpu_id`, so I've removed them.

I find the `value` parameter to be ambiguous so I've tried to make it clearer.

### DIFF
--- a/libsel4vm/src/arch/arm/vgic/vgic.c
+++ b/libsel4vm/src/arch/arm/vgic/vgic.c
@@ -305,7 +305,7 @@ static inline void set_sgi_ppi_pending(struct gic_dist_map *gic_dist, int irq, i
     }
 }
 
-static inline void set_spi_pending(struct gic_dist_map *gic_dist, int irq, int value, int vcpu_id)
+static inline void set_spi_pending(struct gic_dist_map *gic_dist, int irq, int value)
 {
     if (value) {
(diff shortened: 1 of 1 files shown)


axel-h (approved):
Thanks, looks good to me

Ivan-Velickovic:
The further I look at the VGIC code, the more of this kind of thing I see since flags like `-Wunused` aren't passed in when building. I'll stop making further small changes like this and bundle it all into one and hopefully be done with it.

axel-h:
Actually, small changes the are easy to review and merge, are nice. Bigger PRs that refactor a lot tend to take loner.

Ivan-Velickovic:
Okay, I'll keep that in mind then, thanks.

Ivan-Velickovic:
Line 381 had parameter as `bool enable` instead of `bool set_enable` so I changed that.

kent-mcleod:
I think this should be queued for merging after https://github.com/seL4/seL4_projects_libs/pull/28 is resolved as both PRs change the same lines in different, but compatible, ways.

kent-mcleod (approved):

Ivan-Velickovic:
This is ready to go now that #28 has gone through and this PR has been rebased.

lsf37 (approved):

PR #34, opened 2022-03-10
libsel4vm: make macro LR_OF_NEXT a function

axel-h:
The implementation now makes clear this works on `struct lr_of` and asserts for invalid indices to catch potential errors.


libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -192,6 +191,14 @@ struct lr_of {
     bool full;
 };
 
+static inline size_t LR_OF_NEXT(size_t idx)
+{
+    struct lr_of *dummy_lr_of; /* need this to use ARRAY_SIZE() */
+    const size_t max = ARRAY_SIZE(dummy_lr_of->irqs) - 1;
+    assert(idx < max); /* the index must be sane when calling this */
```
Ivan-Velickovic, line 198: Maybe I'm confused but this should be `idx <= max` right?

Ivan-Velickovic, line 198: Maybe I'm confused but this should be `idx <= max` right?

axel-h: Thanks, yes, shame on me. This was `cnt` first, but then I made this `max`.

Ivan-Velickovic: Easy mistake to make :)

PR #36, opened 2022-03-18
libsel4vm: remove obsolete includes

axel-h:
This include file is deprecated since 2016, and it seems there's nothing in there that is needed. Removing the reference here is a preparation to delete this file from the kernel includes completely.

See https://github.com/axel-h/camkes-vm/pull/2 for CI tests.


wom-bat (approved):
This looks good to me.

PR #37, opened 2022-03-23
make LR_OF_NEXT() more robust

axel-h:
Apply defensive programming paradigm and handle invalid indices gracefully in release builds. Reset them to 0 instead of incrementing even further.

### DIFF
--- a/libsel4vm/src/arch/arm/vgic/vgic.c
+++ b/libsel4vm/src/arch/arm/vgic/vgic.c
@@ -183,16 +183,25 @@ struct gic_dist_map {
 #define NUM_LIST_REGS 4
 /* This is a rather abritrary number, increase if needed. */
 #define MAX_IRQ_QUEUE_LEN 64
-#define IRQ_QUEUE_NEXT(_i) (((_i) + 1) & (MAX_IRQ_QUEUE_LEN - 1))
-
-static_assert((MAX_IRQ_QUEUE_LEN & (MAX_IRQ_QUEUE_LEN - 1)) == 0,
-              "IRQ ring buffer size must be power of two");
 
 struct irq_queue {
     struct virq_handle *irqs[MAX_IRQ_QUEUE_LEN]; /* circular buffer */
     size_t head;
     size_t tail;
 };
+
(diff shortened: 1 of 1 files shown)


libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -195,8 +195,13 @@ static inline size_t LR_OF_NEXT(size_t idx)
...
     const size_t idx_max = ARRAY_SIZE(dummy_lr_of->irqs) - 1;
-    assert(idx <= idx_max); /* the index must be sane when calling this */
-    return (idx == idx_max) ? 0 : (idx + 1);
+    /* The index must be sane when calling this, otherwise something seems
+     * broken. In debug builds the assert will stop execution to highlight the
+     * problem. In release builds it's handled gracefully and the index starts
+     * at 0 again - that's be best we can do here.
+     */
+    assert(idx <= idx_max);
+    return (idx <= idx_max) ? (idx + 1) : 0;
```
Ivan-Velickovic, line 204: If you want to return `idx + 1`, only when it would be a valid index into the array, you should be doing `idx < idx_max` right?

kent-mcleod:
I think this should be put on hold until after https://github.com/seL4/seL4_projects_libs/pull/28 is resolved as it looks to be intending to remove this code and it doesn't make sense to review both concurrently.

libsel4vm/src/arch/arm/vgic/vgic.c
axel-h: Yes

hlyytine:
It appears you forgot that the struct name has changed. But what if I just added `struct irq_queue` and forgot to remove `struct lr_of` and that the mistake also eluded the reviewers. That happens sometimes, right (otherwise we wouldn't have obsoleted and unused code anywhere)?  Now this problematic PR would have an opportunity to pass unnoticed and the function you added would be comparing unrelated things.

kent-mcleod:
With the earlier refactoring of the existing LR_OF_NEXT, invalid indices are now already handled by the mask operation in release builds. It doesn't seem that this change is needed anymore.

PR #40, opened 2022-03-28
trivial: use NUM_SGI_VIRQS

axel-h:
Avoid using plain magic numbers.

### DIFF
--- a/libsel4vm/src/arch/arm/vgic/vgic.c
+++ b/libsel4vm/src/arch/arm/vgic/vgic.c
@@ -551,7 +551,7 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
 {
     /* STATE g) */
     struct gic_dist_map *gic_dist = vgic_priv_get_dist(d);
-    if (irq >= 16) {
+    if (irq >= NUM_SGI_VIRQS) {
         DDIST("disabling irq %d\n", irq);
         set_enable(gic_dist, irq, false, vcpu->vcpu_id);
     }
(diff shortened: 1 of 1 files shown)


libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -551,7 +551,7 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
 {
     /* STATE g) */
     struct gic_dist_map *gic_dist = vgic_priv_get_dist(d);
-    if (irq >= 16) {
```
kent-mcleod, line 554: Are you able to add a comment here that this is because 0-15 cannot be disabled.

kent-mcleod (approved):

kent-mcleod (approved):

libsel4vm/src/arch/arm/vgic/vgic.c
axel-h: added a commit on top that adds a comment and also a debug log output

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -551,7 +551,11 @@ static int vgic_dist_disable_irq(struct vgic_dist_device *d, vm_vcpu_t *vcpu, in
 {
     /* STATE g) */
     struct gic_dist_map *gic_dist = vgic_priv_get_dist(d);
-    if (irq >= NUM_SGI_VIRQS) {
+
+    /* SGIs cannot be disabled, such a request is ignored. */
+    if (irq < NUM_SGI_VIRQS) {
+        DDIST("can't disable SGI %d (vCPU %d)\n", irq, vcpu->vcpu_id);
```
axel-h, line 557: @kent-mcleod: should this become a `ZF_LOGW()` or event `ZF_LOGE()`

kent-mcleod: I don't think so.  I recall that for GICv2:
> Whether SGIs are permanently enabled, or can be enabled and disabled by writes to the GICD_ISENABLERn and
> GICD_ICENABLERn, is IMPLEMENTATION DEFINED . 

And then for the particular GIC that is being emulated, it is ignored.  

Guest kernel's can still try and disable all IRQs as part of a startup routine.  seL4 does this in `gic_v2.c`.  So printing a warning would generate a lot of spurious warnings I expect.

PR #41, opened 2022-04-11

feitiantiger:
Error: "vm_ram_touch@guest_ram.c:157 Failed to touch ram region: Not registered RAM region"
 Rootcause: if value(unsigned int) is larger than 0x100000(it is common when the memory of VM allocated exceeds 3G),
 and then left shift 12 bit, the highest bits of 0x1 of value will be lost

 signed-off-by: hanjinglong <hjlallen@qq.com>


axel-h:
Thanks. Indeed, makes sense to cast this before shifting.

PR #42, opened 2022-04-11

feitiantiger:
Error:"vm_ram_touch@guest_ram.c:157
 Failed to touch ram region: Not registered RAM region"
 
 Rootcause:if value(unsigned int) is larger than 0x100000
 (it is common when the memory of VM allocated exceeds 3G),
 and then left shift 12 bit, the highest bits 0x1 of value will be lost.

 Signed-off-by: hanjinglong <hjlallen@qq.com>


feitiantiger:
Need to resubmit

axel-h:
No need to create an new PR, it tracks a branch. So just push to the branch if you have to change something. Force-push also works.

feitiantiger:
Sorry, first submission, not familiar with the process，new PR has been submit

PR #43, opened 2022-04-12

feitiantiger:
Error: "vm_ram_touch@guest_ram.c:157 Failed to
touch ram region: Not registered RAM region".

Rootcause: if the value(unsigned int) is larger than 0x100000
(it is common when the memory of VM allocated exceeds 3G),
and then left shift 12 bit, the highest bits 0x1 of value will be lost.

Signed-off-by: hanjinglong <hjlallen@qq.com>

### DIFF
--- a/libsel4vmmplatsupport/src/drivers/virtio_emul.c
+++ b/libsel4vmmplatsupport/src/drivers/virtio_emul.c
@@ -86,7 +86,7 @@ static int emul_io_out(virtio_emul_t *emul, unsigned int offset, unsigned int si
         assert(size == 4);
         int queue = emul->virtq.queue;
         emul->virtq.queue_pfn[queue] = value;
-        vring_init(&emul->virtq.vring[queue], emul->virtq.queue_size[queue], (void *)(uintptr_t)(value << 12),
+        vring_init(&emul->virtq.vring[queue], emul->virtq.queue_size[queue], (void *)((uintptr_t)value << 12),
                    VIRTIO_PCI_VRING_ALIGN);
         break;
     }


axel-h:
Thanks. Indeed, makes sense to cast this before shifting.

libsel4vmmplatsupport/src/drivers/virtio_emul.c
```diff
@@ -86,7 +86,7 @@ static int emul_io_out(virtio_emul_t *emul, unsigned int offset, unsigned int si
         assert(size == 4);
         int queue = emul->virtq.queue;
         emul->virtq.queue_pfn[queue] = value;
-        vring_init(&emul->virtq.vring[queue], emul->virtq.queue_size[queue], (void *)(uintptr_t)(value << 12),
+        vring_init(&emul->virtq.vring[queue], emul->virtq.queue_size[queue], (void *)((uintptr_t)value << 12),
```
axel-h, line 89: Maybe adding a check (or at least an assert) would be good to ensure the shift by 12 does not overflow. For 64-bit this seem quite unlikely, but for 32-bit it might still happen.

feitiantiger: if (4 = sizeof(uintptr_t)) {
    assert(value < BIT(20));
}

Your mean like this?

feitiantiger: Well，although it lookes like generic, the value notified by Host with iowrite32 is unsigned int type, the valid range is less than BIT(32) even for 64-bit, so value >= BIT((8 * sizeof(uintptr_t)) - 12) is never ture

wom-bat (approved):
Looks good.

libsel4vmmplatsupport/src/drivers/virtio_emul.c
wom-bat: I don't think the assert is necessary, given the sizes and types of the operands.  It may be worth scanning the code for other cases where the cast is in the wrong place.

PR #46, opened 2022-04-28
trivial: return ret when generating vcpu node

chrisguikema:
Signed-off-by: Chris Guikema <chris.guikema@dornerworks.com>


axel-h (approved):
looks ok to me, this seems to fix the error propagation.

PR #52, opened 2022-05-11
libsel4vm: merge definition of virq_handle_t

axel-h:


Ivan-Velickovic:
The typedef is only used in one place in the code, everywhere else uses `struct virq_handle *`, perhaps it'd be better to just get rid of the typedef?

Ivan-Velickovic:
The typedef is only used in one place in the code, everywhere else uses `struct virq_handle *`, perhaps it'd be better to just get rid of the typedef?

axel-h:
That's a point. For some reason this code seems to prefer the explicit struct usage over typedefs...

libsel4vm/src/arch/arm/vgic/vgic.c
```diff
@@ -121,13 +121,11 @@
...
 #define IRQ_BIT(irq) (1U << ((irq) % 32))
 
-struct virq_handle {
+typedef struct virq_handle {
     int virq;
     irq_ack_fn_t ack;
     void *token;
-};
-
-typedef struct virq_handle *virq_handle_t;
```
kent-mcleod, line 130: This typedef isn't allowed by the style guide: https://docs.sel4.systems/processes/style-guide.html#naming-of-symbols-variables-types-and-type-aliases-typedef.

axel-h: goot point, weill remove this from the code then in a separate PR.

PR #61, opened 2022-06-27
libsel4vm: leave IPC details in VM code

axel-h:
- keep vgic code agnostic of IPC details. Currently it is a bit odd how thing are split between vm.c and vgic.c, having vgic agnostic of the notification details allows better reusing/testing the code.
- inline handle_vgic_maintenance(), simplified the code
- use ZF_LOGE() instead of ZF_LOGF(), caller stops on error anyway.

### DIFF
--- a/libsel4vm/src/arch/arm/vgic/vgic.c
+++ b/libsel4vm/src/arch/arm/vgic/vgic.c
@@ -489,30 +489,6 @@ static int vgic_vcpu_load_list_reg(vgic_t *vgic, vm_vcpu_t *vcpu, int idx, struc
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


kent-mcleod:
> keep vgic code agnostic of IPC details. Currently it is a bit odd how thing are split between vm.c and vgic.c, having vgic agnostic of the notification details allows better reusing/testing the code.

I'm not a fan of this change.  It doesn't seem to get us anything and adds more gic code into the vm.c file when they've been intentionally separated.  The vgic code isn't supposed to be agnostic of IPC details as the message format of the maintenance fault message is allowed to be GIC model specific based on the hardware capabilities.

PR #64, opened 2022-07-04
fix active_clr0 access 

axel-h:
Reading `active0` to write `active_clr0` looks wrong somehow.


kent-mcleod:
> Reading `active0` to write `active_clr0` looks wrong somehow.

Yea, but the driver doesn't really emulate the active state for each IRQ at the moment. It needs to be able to handle a guest clearing the active state for each IRQ on startup. And otherwise it currently assumes that a guest doesn't read or write the active state otherwise.

kent-mcleod (approved):

PR #58, opened 2022-06-16
Updated: GICv3 emulation

kent-mcleod:
This PR adds an emulated GIC500 (GICv3) to libsel4vmm as a virtual device. 

This branch is based off of an earlier PR, https://github.com/seL4/seL4_projects_libs/pull/8, by @pingerino which already included commits pulled together from a couple authors.
This has been tested with the imx8mm-evk platform to run a Linux guest VM with several pass-through peripherals. 

Marking the PR as a draft for now, as some minor cleanup is still needed.

PR #68, opened 2022-07-10
GICv3 emulation preparation PR

kent-mcleod:
These changes are the first few commits from https://github.com/seL4/seL4_projects_libs/pull/58/ which reorganize the vgic code to allow adding support for GICv3.

For reviewing, these changes shouldn't be changing any behavior of the current implementation and are only supposed to be moving code around with a few function prototype changes, eg `static` -> `static inline`.


libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -0,0 +1,300 @@
...
+
+static struct vgic_dist_device *vgic_dist;
+
+static inline struct gic_dist_map *vgic_priv_get_dist(struct vgic_dist_device *d)
+{
+    assert(d);
+    assert(d->priv);
+
+    vgic_t *vgic = vgic_device_get_vgic(d);
+    return priv_get_dist(vgic->dist);
```
axel-h, line 70:
```suggestion
    return vgic->dist;
```
This seems a merge artifact.

kent-mcleod: Thanks for spotting. I moved the priv_get_dist changes into a later commit so that the move commits didn't also modify the code.

libsel4vm/src/arch/arm/vgic/virq.h
```diff
@@ -0,0 +1,225 @@
...
+}
+
+static inline void virq_init(virq_handle_t virq, int irq, irq_ack_fn_t ack_fn, void *token)
+{
+    virq->virq = irq;
+    virq->token = token;
+    virq->ack = ack_fn;
+}
+
+static inline struct vgic *vgic_device_get_vgic(struct vgic_dist_device *d)
```
axel-h, line 159: This function does not really belong here, it would fit better in `vgic_v2.c` even if this duplicates code for GICv3 then. I have a branch where `struct vgic_dist_device` is dropped completely and this is all part of the `struct vgic` then, but have to check how this would work with the GICv3 code then.

kent-mcleod: What you say makes sense about the function staying in vgic_v2.c, I need to check with the old version whether it had it in there or whether I accidentally pulled it in during a rebase. 

I'm not sure whether dropping vgic_dist_device would make the code cleaner and it'd be better to wait until v3 code is merged before we consider IMO.

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -0,0 +1,300 @@
...
+    int err = virq_add(vcpu, vgic, virq_data);
+    if (err) {
+        free(virq_data);
+        return -1;
+    }
+
+    return 0;
+}
+
+int vm_inject_irq(vm_vcpu_t *vcpu, int irq)
```
axel-h, line 179: It seem this function remains the same in GICv2 and GICv3. Thus I think it would make sense to keep thios in a common file and avoid code duplication. If there are differences in the future, we can still split this up,

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -0,0 +1,300 @@
...
+    int err = vm_map_reservation(vm, vgic_vcpu_reservation, vgic_vcpu_iterator, (void *)vm);
+    if (err) {
+        free(vgic_dist->priv);
+        return -1;
+    }
+
+    return 0;
+}
+
+int vm_vgic_maintenance_handler(vm_vcpu_t *vcpu)
```
axel-h, line 279: Same here, this function and alsdo `handle_vgic_maintenance()` are the same in GICv2 and GIV3, so I'd keep this in a common file for now.

Ivan-Velickovic: They're slightly different since GICv3 uses group 1 while GICv2 uses group 0.

Ivan-Velickovic: (only `handle_vgic_maintenance` differs)

axel-h: True, but that could be a constant then. My main point at the moment it trying to have GICv3 support with out renaming the source file and moving code, so the change are easier to maintain.

pingerino (approved):

axel-h (approved):
I not too happy with some things in the code, but it the greater goal is adding some GICv3 support finally, then I wont block this.

libsel4vm/src/arch/arm/vgic/virq.h
kent-mcleod: I added a new commit to address this: https://github.com/seL4/seL4_projects_libs/pull/68/commits/0ec24c1c8c13a73a7441bd9290dbab5a46c71424.  I think it's a bit reactive, but still reasonable to fix it this way.

pingerino (approved):

PR #71, opened 2022-07-14
libsel4vm: remove unused function is_active()

axel-h:
This was removed in c3d56860 already, but the refactoring in 6c0c5d22 brought it back accidentally.


kent-mcleod:
> This was removed in [c3d5686](https://github.com/seL4/seL4_projects_libs/commit/c3d5686032f94232a5a04c50699d4a6d4f725c10) already, but the refactoring in [6c0c5d2](https://github.com/seL4/seL4_projects_libs/commit/6c0c5d226ac7b3b8a54bfd520bab5cabbcc013ce) brought it back accidentally.

This isn't the case. That commit removed `set_active`, `is_active` was still in vgic.c and was moved.  If you look at the [move commit](https://github.com/seL4/seL4_projects_libs/commit/6c0c5d226ac7b3b8a54bfd520bab5cabbcc013ce), it shouldn't add or remove any code other than license headers and some includes.

axel-h:
Good catch, my bad. But for `is_active()` I can't see any usage either, seems this could be dropped then also.

kent-mcleod:
> Good catch, my bad. But for `is_active()` I can't see any usage either, seems this could be dropped then also.

Yea, if we thought it was worth removing set_active then it should apply the same way for is_active. We don't do proper active handling anyway.  This is because the kernel doesn't provide the information.  To implement the active registers, you would need to look inside the current list registers for the CPU interface to see which interrupts have been moved from pending to active or activepending by the guest, but I don't believe that the kernel provides a way for the vmm to read the values of the list registers, only sending a maintenance interrupt when the interrupt is deactivated. For completeness we should handle this, but in practice it's not needed unless we want to allow the guest to be able to suspend and resume itself.

PR #76, opened 2022-10-20
fdt_gen: Return error values correctly

chrisguikema:
This makes sure if generate_psci_node() returns an error other than FDT_ERR_EXISTS that it does not get ignored by
fdt_generate_plat_vcpu_node().

### DIFF
--- a/libsel4vmmplatsupport/src/arch/arm/guest_vcpu_util.c
+++ b/libsel4vmmplatsupport/src/arch/arm/guest_vcpu_util.c
@@ -71,7 +71,11 @@ static int generate_psci_node(void *fdt, int root_offset)
 {
     int psci_node = fdt_add_subnode(fdt, root_offset, "psci");
     if (psci_node < 0) {
-        return psci_node;
+        if (-FDT_ERR_EXISTS == psci_node) {
+            return 0;
+        } else {
+            return psci_node;
+        }
     }
     FDT_OP(fdt_appendprop_u32(fdt, psci_node, "cpu_off", 0x84000002));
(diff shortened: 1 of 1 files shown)


libsel4vmmplatsupport/src/arch/arm/guest_vcpu_util.c
```diff
@@ -87,6 +91,7 @@ int fdt_generate_plat_vcpu_node(vm_t *vm, void *fdt)
     int root_offset = fdt_path_offset(fdt, "/");
     int cpu_node = fdt_add_subnode(fdt, root_offset, "cpus");
     if (cpu_node < 0) {
+        ZF_LOGE("Failed to create cpu node");
         return cpu_node;
     }
     FDT_OP(fdt_appendprop_u32(fdt, cpu_node, "#address-cells", 0x1));
```
axel-h, line 97: I lack some background about the code, so I wonder what happens now if `cpu_node` is '0`? Doesn't this expect to get some handle here? Or can this handle `0` then also?

axel-h: What I wonder is, does it make sense to pass `0` there now of should we also check for `if (0 == cpu_node) {` now also and don't try to do anything at all then - do we don't run into error that are expected anyway, which could be confusing.

kent-mcleod: If `fdt_add_subnode`'s return value is >= 0, then it is supposed to be interpreted as an offset to the newly created node.

kent-mcleod (approved):

PR #77, opened 2022-10-20
libsel4vm: change vcpu thread naming scheme

chrisguikema:
This makes it easier to differentiate between the vcpu threads when debugging and/or testing

### DIFF
--- a/libsel4vm/src/arch/arm/boot.c
+++ b/libsel4vm/src/arch/arm/boot.c
@@ -19,6 +19,7 @@
 
 #include <sel4vm/boot.h>
 #include <sel4vm/guest_vm.h>
+#include <sel4vm/guest_vm_util.h>
 #include <sel4vm/arch/guest_arm_context.h>
 
 #include "arm_vm.h"
@@ -109,5 +110,12 @@ int vm_create_vcpu_arch(vm_t *vm, vm_vcpu_t *vcpu)
         err = -1;
     }
 #endif /* CONFIG_MAX_NUM_NODES > 1 */
+
+#ifdef CONFIG_DEBUG_BUILD
+    char vcpu_name[32];
+    snprintf(vcpu_name, sizeof(vcpu_name), "%s:%d", vm->vm_name, vcpu->vcpu_id);
+    seL4_DebugNameThread(vm_get_vcpu_tcb(vcpu), vcpu_name);
+#endif
+
     return err;
 }


libsel4vm/src/arch/arm/boot.c
```diff
@@ -109,5 +114,9 @@ int vm_create_vcpu_arch(vm_t *vm, vm_vcpu_t *vcpu)
         err = -1;
     }
 #endif /* CONFIG_MAX_NUM_NODES > 1 */
+
+    snprintf(vcpu_name, VCPU_NAME_SIZE, "%s:Core%d", vm->vm_name, vcpu->vcpu_id);
+    NAME_THREAD(vm_get_vcpu_tcb(vcpu), vcpu_name);
```
axel-h, line 119: I'd make all this a bit more compact and have just one place in the code that changes. That also avoids unused variable warnings for release builds. Also, having just 5 chars for the VM name seems not enough, so allowing 32 chars and using "%s:%d"` might leave more space there:
```
#ifdef CONFIG_DEBUG_BUILD
    char vcpu_name[32]; 
    snprintf(vcpu_name, sizeof(vcpu_name), "%s:%d", vm->vm_name, vcpu->vcpu_id);
    seL4_DebugNameThread(vm_get_vcpu_tcb(vcpu), vcpu_name);
#endif
```

Seem we also have an `asprintf()` implementation, so it could be this flexible version:
```
#ifdef CONFIG_DEBUG_BUILD
    char *vcpu_name;
    asprintf(&vcpu_name, "%s:%d", vm->vm_name, vcpu->vcpu_id);
    assert(vcpu_name); 
    seL4_DebugNameThread(vm_get_vcpu_tcb(vcpu), vcpu_name);
    free(vcpu_name):
#endif
```
Actually, we could also print into the IPC buffer directly, the internal strcpy() should be a nop then.

#ifdef CONFIG_DEBUG_BUILD
    /* seL4_DebugNameThread() passes the name in the IPC buffer, so we can put it there directly)
    char *vcpu_name = seL4_GetIPCBuffer()->msg;
    snprintf(vcpu_name, seL4_MsgMaxLength, "%s:%d", vm->vm_name, vcpu->vcpu_id);
    seL4_DebugNameThread(vm_get_vcpu_tcb(vcpu), vcpu_name);
#endif

Which make we wonder if the seL4 API should get a `seL4_DebugFmtNameThread(fmt, ...)` for convenience reasons?

chrisguikema: I prefer to use `snprintf` where possible, so I implemented to first version. I do like the idea of a convenient DebugFmtNameThread convenience function though!

kent-mcleod (approved):

PR #78, opened 2022-10-21
Add zynqmp support to libsel4vm

chrisguikema:
I know its not a supported platform yet, but this should be able to be merged and it will be ready when the camkes-vm officially supports the board.

### DIFF
--- a/libsel4dma/src/dma.c
+++ b/libsel4dma/src/dma.c
@@ -28,6 +28,8 @@
 #define DMA_MINALIGN_BYTES 32
 #elif defined(CONFIG_PLAT_EXYNOS5)
 #define DMA_MINALIGN_BYTES 32
+#elif defined(CONFIG_PLAT_ZYNQMP)
+#define DMA_MINALIGN_BYTES 32
 #else
 #warning Unknown platform. DMA alignment defaulting to 32 bytes.
 #define DMA_MINALIGN_BYTES 32
--- a/libsel4vm/src/arch/arm/vgic/gicv2.h
+++ b/libsel4vm/src/arch/arm/vgic/gicv2.h
@@ -22,6 +22,8 @@
 #define GIC_PADDR   0x8000000
 #elif defined(CONFIG_PLAT_ODROIDC2)
 #define GIC_PADDR   0xc4300000
+#elif defined(CONFIG_PLAT_ZYNQMP)
+#define GIC_PADDR   0xf9000000
 #else
 #error "Unsupported platform for GIC"
 #endif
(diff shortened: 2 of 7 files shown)


libsel4vmmplatsupport/include/sel4vmmplatsupport/ringbuf.h
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2018, DornerWorks
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <stddef.h>
```
axel-h, line 7: Add `#pragma once`

libsel4vmmplatsupport/src/ringbuf.c
```diff
@@ -0,0 +1,57 @@
+/*
```
axel-h, line 1: I wonder if `libsel4vmmplatsupport` is the best place for such a lib. We already have https://github.com/seL4/projects_libs/tree/master/libringbuffer, seems the new one is a bit more advanced. Would it make sense to add it there?

kent-mcleod: Or alternatively, keep this implementation private to the zynqmp/devices/vuart.c for now

libsel4vmmplatsupport/src/plat/zynqmp/devices/vuart.c
```diff
@@ -0,0 +1,397 @@
+/*
```
kent-mcleod, line 1: Instead of vuart.c, would it be easier to name this file based on which device it is emulating? eg virt-cdns-uart-r1p12.c or whatever the device compatibility string is for the device that this emulates?

kent-mcleod:
> I know its not a supported platform yet, but this should be able to be merged and it will be ready when the camkes-vm officially supports the board.

Is camkes-vm support blocked on SMC calls?

chrisguikema:
> Is camkes-vm support blocked on SMC calls?

Essentially yes, but technically we could release support for RTEMS. Linux has far too many SMC calls that must be emulated.

libsel4vmmplatsupport/src/plat/zynqmp/devices/vuart.c
chrisguikema: That would be good behavior. I'm honestly thinking at this point that the `vuart` be refactored out of this PR and into its own, since the vuart isn't technically necessary (passthrough serial and virtio-console both work well)

kent-mcleod (approved):

PR #79, opened 2022-10-24
libsel4vm: add vuart for zynqmp

chrisguikema:
This implementation has made several design decisions:
    * No concept of an "active" uart
    * The vuart TEMPTY interrupt status flag is kept asserted
    * There is a predefined console
    * Default serial port is UART0

The TEMPTY interrupt status flag is always asserted because it is set when writing the TX FIFO, but cleared when the interrupt is acknowledged. This causes a problem if a write to the TX FIFO occurs in the guest's interrupt handler, because the interrupt could be acknowledged after writing to the TX FIFO.

The vuart device immediately flushes all writes to the TX FIFO, so it is always empty. This commit will keep the TEMPTY flag set indefinitely to more closely emulate real hardware behavior.

Signed-off-by: Chris Guikema <chris.guikema@dornerworks.com>

### DIFF
--- a/libsel4vmmplatsupport/CMakeLists.txt
+++ b/libsel4vmmplatsupport/CMakeLists.txt
@@ -61,6 +61,7 @@ target_link_libraries(
     fdt
     fdtgen
     satadrivers
+    ringbuffer
     sel4_autoconf
     sel4vm_Config
     usbdrivers_Config
--- a/libsel4vmmplatsupport/plat_include/zynqmp/sel4vmmplatsupport/plat/device_map.h
+++ b/libsel4vmmplatsupport/plat_include/zynqmp/sel4vmmplatsupport/plat/device_map.h
@@ -5,3 +5,9 @@
  */
 
 #pragma once
+
+/* UART */
+#define UART0_PADDR           0xFF000000
+#define UART1_PADDR           0xFF010000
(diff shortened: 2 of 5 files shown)


libsel4vmmplatsupport/src/plat/zynqmp/devices/vuart.c
```diff
@@ -0,0 +1,398 @@
+/*
```
kent-mcleod, line 1: Can this device name be updated to indicate which device model it is emulating?

chrisguikema: sure thing!

chrisguikema:
@kent-mcleod is everything good here?

kent-mcleod (approved):

kent-mcleod:
> @kent-mcleod is everything good here?

All of the zynqmp support doesn't have CI for now until the SMC rfc is progressed, so I've only really reviewed that this code makes sense structurally and doesn't impact other platforms.

chrisguikema:
Thanks! That was my hope, just to reduce the amount of backlog we have internally. My plan is to create PRs that release a fully functional example once the SMC RFC is progressed, and that may require a few tweaks here and there

PR #80, opened 2022-10-25
guest_vspace: don't map object w/o rights to DMA

chrisguikema:
Objects that can't be read from or written to don't have a point in being in a DMA space.

### DIFF
--- a/libsel4vm/src/guest_vspace.c
+++ b/libsel4vm/src/guest_vspace.c
@@ -43,6 +43,14 @@ static int guest_vspace_map(vspace_t *vspace, seL4_CPtr cap, void *vaddr, seL4_C
     }
 
 #if defined(CONFIG_TK1_SMMU) || defined(CONFIG_IOMMU)
+    /* If an object doesn't have Read or Write access, there isn't a reason to put
+     * it in DMA space
+     */
+    if (rights.words[0] == seL4_NoRights.words[0]) {
+        ZF_LOGW("Don't put object without RW access into DMA space");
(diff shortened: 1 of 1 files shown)


libsel4vm/src/guest_vspace.c
```diff
@@ -43,6 +43,14 @@ static int guest_vspace_map(vspace_t *vspace, seL4_CPtr cap, void *vaddr, seL4_C
     }
 
 #if defined(CONFIG_TK1_SMMU) || defined(CONFIG_IOMMU)
+    /* If an object doesn't have Read or Write access, there isn't a reason to put
+     * it in DMA space
+     */
+    if (rights.words[0] == seL4_NoRights.words[0]) {
```
kent-mcleod, line 49: Is this solving some issue that was encountered? The same argument could be made to creating mappings without rights to the VMM memspace too, but its something that we don't restrict because it doesn't typically create a problem, and maybe the VMM could be trying to implement on-demand paging or something.

chrisguikema: I'm not sure... 

This was part of our backlog that we're trying to work through. If I remember correctly, we encountered some issue, but that was with the old SMMU driver that utilized iospaces, so this may not even be necessary. I figured i'd at least push this up for review and discussion, but we can always mark it as not needed.

chrisguikema: If only on-demand DMA mapping would work...oh well!

I'll close this PR.

PR #82, opened 2022-10-26
libsel4vm, arm: rework load_image

chrisguikema:
Previously, load_image used a 1MB buffer from the heap to read the guest images into. Free-ing this buffer would cause a warning to appear.

"libsel4muslcsys: Error attempting syscall 215"

This commit reworks the load_image function to use the touch callback to read the file from the file server, thus eliminating the need for malloc-ing and free-ing a large buffer.

This fixes the issue discussed in https://github.com/seL4/seL4_libs/pull/70/

### DIFF
--- a/libsel4vmmplatsupport/src/arch/arm/guest_image.c
+++ b/libsel4vmmplatsupport/src/arch/arm/guest_image.c
@@ -134,7 +134,15 @@ static int get_guest_image_type(const char *image_name, enum img_type *image_typ
 
 static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t size, size_t offset, void *cookie)
 {
-    memcpy(vaddr, cookie + offset, size);
+    int fd = *((int *)cookie);
+
+    /* Load the image */
+    size_t len = read(fd, vaddr, size);
+    if (len != size) {
+        ZF_LOGE("Bytes read from the file server (%d) don't match expected length (%d)", len, size);
+        return -1;
+    }
+
     if (config_set(CONFIG_PLAT_TX1) || config_set(CONFIG_PLAT_TX2)) {
         seL4_CPtr cap = vspace_get_cap(&vm->mem.vmm_vspace, vaddr);
         if (cap == seL4_CapNull) {
@@ -151,45 +159,31 @@ static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t si
 static int load_image(vm_t *vm, const char *image_name, uintptr_t load_addr,  size_t *resulting_image_size)
 {
     int fd;
-    size_t len;
     int error;
+
     fd = open(image_name, 0);
     if (fd == -1) {
         ZF_LOGE("Error: Unable to find image \'%s\'", image_name);
         return -1;
     }
 
-    /* Try and load the image 1MiB at a time. Reduce the size by half if
-     * there isn't enough memory available.  The total loading time may be
-     * faster if a larger buffer is used as it allows for more batching.
-     */
-    size_t read_size = BIT(20);
-    char *buf = malloc(read_size);
-    while (buf == NULL) {
-        read_size /= 2;
-        if (read_size < 4096) {
-            ZF_LOGE("Not enough memory for copy buffer");
-        }
-        buf = malloc(read_size);
+    size_t file_size = lseek(fd, 0, SEEK_END);
+    lseek(fd, 0, SEEK_SET);
+
+    if (0 == file_size) {
+        ZF_LOGE("Error: \'%s\' has zero size", image_name);
+        return -1;
     }
-    size_t offset = 0;
-    while (1) {
-        /* Load the image */
-        len = read(fd, buf, read_size);
-        if (!len) {
-            break;
-        }
-        vm_ram_mark_allocated(vm, load_addr + offset, len);
-        error = vm_ram_touch(vm, load_addr + offset, len, guest_write_address, (void *)buf);
-        if (error) {
-            ZF_LOGE("Error: Failed to load \'%s\'", image_name);
-            close(fd);
-            return -1;
-        }
-        offset += len;
+
+    vm_ram_mark_allocated(vm, load_addr, ROUND_UP(file_size, PAGE_SIZE_4K));
+    error = vm_ram_touch(vm, load_addr, file_size, guest_write_address, (void *)&fd);
+    if (error) {
+        ZF_LOGE("Error: Failed to load \'%s\'", image_name);
+        close(fd);
+        return -1;
     }
-    free(buf);
-    *resulting_image_size = offset;
+
+    *resulting_image_size = file_size;
     close(fd);
     return 0;
 }


libsel4vmmplatsupport/src/arch/arm/guest_image.c
```diff
@@ -145,52 +152,57 @@ static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t si
...
-        len = read(fd, buf, read_size);
-        if (!len) {
-            break;
-        }
-        vm_ram_mark_allocated(vm, load_addr + offset, len);
-        error = vm_ram_touch(vm, load_addr + offset, len, guest_write_address, (void *)buf);
-        if (error) {
+    while (offset < file_size) {
+        /* Will return bytes copied to guest memory, or -1 on error */
+        len = vm_ram_touch(vm, load_addr + offset, PAGE_SIZE_4K, guest_write_address, (void *)fd);
```
kent-mcleod, line 196: I recall that this function is supposed to support large lengths and it then calls the callback several times, usually one for each page. It should be possible to install a ram device with the largest pages possible and then have this run once for each page?

chrisguikema: Yes, you're right

chrisguikema:
@kent-mcleod any more thoughts on this one?

libsel4vmmplatsupport/src/arch/arm/guest_image.c
```diff
@@ -148,49 +155,48 @@ static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t si
...
+    fclose(file);
+
+    return size;
+}
+
 static int load_image(vm_t *vm, const char *image_name, uintptr_t load_addr,  size_t *resulting_image_size)
 {
-    int fd;
-    size_t len;
+    int fd[1];
```
axel-h, line 175: What cant this remain `int fd;`, seems we make it a 1-element array now for no reason?

chrisguikema: You're right, it can just stay as an int. While I was putting this patch together I was playing around with the idea of passing some other information to the callback, but it wasn't necessary.

libsel4vmmplatsupport/src/arch/arm/guest_image.c
```diff
@@ -148,48 +155,47 @@ static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t si
...
+    return size;
+}
+
 static int load_image(vm_t *vm, const char *image_name, uintptr_t load_addr,  size_t *resulting_image_size)
 {
     int fd;
-    size_t len;
     int error;
+
+    size_t file_size = get_file_size(image_name);
```
axel-h, line 178: Do we need this helper of can we use `stat(fd, &stat_data)` and then get the file size from `stat_data.st_size`?

chrisguikema: That doesn't work.

This block:
```
    struct stat buf;
    stat(image_name, &buf);

    ZF_LOGE("%d | %d", file_size, buf.st_size);
```

Gives this output:
```
libsel4muslcsys: Error attempting syscall 79
load_image@guest_image.c:195 15405568 | 0
Loading Initrd: 'linux-initrd'
libsel4muslcsys: Error attempting syscall 79
load_image@guest_image.c:195 912857 | 0
```

libsel4vmmplatsupport/src/arch/arm/guest_image.c
```diff
@@ -134,7 +134,14 @@ static int get_guest_image_type(const char *image_name, enum img_type *image_typ
 
 static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t size, size_t offset, void *cookie)
 {
-    memcpy(vaddr, cookie + offset, size);
+    int *fd = (int *)cookie;
+
+    /* Load the image */
+    size_t len = read(fd[0], vaddr, PAGE_SIZE_4K);
```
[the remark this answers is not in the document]
chrisguikema: This is a callback from `vm_ram_touch`, which allocates guest memory equal to the file size at the specified chunk. So if `len != size`, then we know we aren't loading the entire file, and we should return an error.

libsel4vmmplatsupport/src/arch/arm/guest_image.c
chrisguikema: That doesn't work either.

I can get things working w/o the helper function by using `fread`:

```
static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t size, size_t offset, void *cookie)
{
    FILE *file = (FILE *)cookie;

    /* Load the image */
    int err = fread(vaddr, size, 1, file);
    if (err != 1) {
        ZF_LOGE("fread failed unexpectedly");
        return -1;
    }

    if (config_set(CONFIG_PLAT_TX1) || config_set(CONFIG_PLAT_TX2)) {
        seL4_CPtr cap = vspace_get_cap(&vm->mem.vmm_vspace, vaddr);
        if (cap == seL4_CapNull) {
            /* Not sure how we would get here, something has gone pretty wrong */
            ZF_LOGE("Failed to get vmm cap for vaddr: %p", vaddr);
            return -1;
        }
        int error = seL4_ARM_Page_CleanInvalidate_Data(cap, 0, PAGE_SIZE_4K);
        ZF_LOGF_IFERR(error, "seL4_ARM_Page_CleanInvalidate_Data failed");
    }
    return 0;
}

static int load_image(vm_t *vm, const char *image_name, uintptr_t load_addr,  size_t *resulting_image_size)
{
    int error;

    FILE *file = fopen(image_name, "r");
    if (!file) {
        ZF_LOGE("Error: Unable to find image \'%s\'", image_name);
        return -1;
    }

    fseek(file, 0, SEEK_END);
    size_t file_size = ftell(file);
    fseek(file, 0, SEEK_SET);

    if (0 == file_size) {
        ZF_LOGE("Error: \'%s\' has zero size", image_name);
        return -1;
    }

    vm_ram_mark_allocated(vm, load_addr, ROUND_UP(file_size, PAGE_SIZE_4K));
    error = vm_ram_touch(vm, load_addr, file_size, guest_write_address, (void *)file);
    if (error) {
        ZF_LOGE("Error: Failed to load \'%s\'", image_name);
        fclose(file);
        return -1;
    }

    *resulting_image_size = file_size;
    fclose(file);
    return 0;
}

@axel-h @kent-mcleod do you guys like that method better? Or should we stick with using the file descriptor / open method?
```

libsel4vmmplatsupport/src/arch/arm/guest_image.c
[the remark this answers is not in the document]
chrisguikema: Oh, and you were right that we can just use `size`.

libsel4vmmplatsupport/src/arch/arm/guest_image.c
axel-h: On a naive view, switching from the bare metal `open()` to the more file specific `fopen()` would be a generic refactoring that could happen first, as it seems this is the proper API-familiy to use for file I/O.

libsel4vmmplatsupport/src/arch/arm/guest_image.c
```diff
@@ -134,7 +134,14 @@ static int get_guest_image_type(const char *image_name, enum img_type *image_typ
 
 static int guest_write_address(vm_t *vm, uintptr_t paddr, void *vaddr, size_t size, size_t offset, void *cookie)
 {
-    memcpy(vaddr, cookie + offset, size);
+    int *fd = (int *)cookie;
```
axel-h, line 137:
```suggestion
    int fd = *((int *)cookie);
```
Let's unwrap this directly, then we can use `fd` everywhere and not `fd[0]`. 
Or actually, it might be the simplest thing f we pass the value of `fd` directly in the cookie?

libsel4vmmplatsupport/src/arch/arm/guest_image.c
chrisguikema: In that case I'll push the `fopen` version. Using the `file` pointer as a cookie is also more straight forward.

kent-mcleod: > OK, thanks for trying. Out of curiosity, does this work with the file descriptor
> 
> ```
>     seek(fd, 0, SEEK_END);
>     size_t size = tell(fd);
>     seek(fd, 0, SEEK_SET);
> ```

@chrisguikema, what was the issue with this approach? `lseek(fd, 0, SEEK_END);` should directly return the filesize in the return value. Then `lseek(fd, 0, SEEK_SET);` should reset it. This is what the fseek and ftell functions are using internally.

It's probably unsafe to use the file streaming layer because muslc assumes extra OS functionality that isn't implemented by sel4muslcsys. It also adds an extra layer of buffering which we don't use.

chrisguikema:
```
    size_t size = lseek(fd, 0, SEEK_END);
    lseek(fd, 0, SEEK_SET);
```

Using `lseek` works. When using `seek` and `tell` there were linker issues. I can update to not use the helper in this case.

libsel4vmmplatsupport/src/arch/arm/guest_image.c
chrisguikema: Updated to grab the contents of the cookie as `fd`.

kent-mcleod (approved):

axel-h (approved):
Looks good to me now. Nice work!

chrisguikema:
Thanks for the review, guys. Always like it when a piece of code gets cleaned up like this. And I can't tell you how many customers we've had that are concerned about the syscall 215 issue

PR #83, opened 2022-10-27
libsel4vm, vgic: generate dist and cpu values

chrisguikema:
This commit adds an "fdt_ori" and "gic_node" parameter to the vm struct, which allows the library to find the GIC_DIST and GIC_CPU register addresses required for emulating the vgic.


chrisguikema:
This isn't perfect, and requires a change in `camkes-vm` to set the path and pointers. It could also use a check to handle address-cells and size-cells properly. I just wanted to put it out for review and discussion, and to show its possible to grab the values from the device tree. I'm open to other and better ways to do this.

chrisguikema:
Test with: https://github.com/seL4/camkes-vm/pull/52

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -292,7 +298,7 @@ int vm_vgic_maintenance_handler(vm_vcpu_t *vcpu)
 }
 
 const struct vgic_dist_device dev_vgic_dist = {
-    .pstart = GIC_DIST_PADDR,
+    .pstart = 0,
```
axel-h, line 301: If this structure is a const, but has only dummy data, it feels a bit that it is obsolete now. Can't we remove it?

libsel4vm/src/arch/arm/guest_irq_controller.c
```diff
@@ -6,11 +6,56 @@
 
 #include "vgic/vgic.h"
 
+#include <libfdt.h>
+
+#define DIST_OFFSET 0
+#define CPU_OFFSET  16
+#define VCPU_OFFSET 48
```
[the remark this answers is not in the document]
chrisguikema: I ended up reworking to reference the size-cells and address-cells. Then the offsets are calculated based on the standard order.

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -58,7 +58,8 @@
 #include "gicv2.h"
 #include "vdist.h"
 
-
+static uintptr_t vcpu_addr;
+static uintptr_t cpu_addr;
```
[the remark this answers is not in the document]
chrisguikema: At this point, yes, it feels like this could be reworked to avoid the globals. Since you've already done some work with that in the PR you referenced, do you want to handle that?

libsel4vm/src/arch/arm/vgic/vgic_v2.c
chrisguikema: I'd hope so! I agree its kind of pointless now.

libsel4vm/src/arch/arm/vgic/vgic_v2.c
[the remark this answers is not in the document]
axel-h: I'd try to rebase my PR on this then.

libsel4vm/src/arch/arm/vgic/vgicv2_defs.h
```diff
@@ -68,3 +68,10 @@
 #define GIC_DIST_SGI_CPU_TARGET_LIST_MASK       0xFF << GIC_DIST_SGI_CPU_TARGET_LIST_SHIFT
 
 #define GIC_DIST_SGI_INTID_MASK                 0xF
+
+/* Contants for getting registers from Device Tree */
+#define NUM_GIC_REGS        4
+#define GIC_DIST_ENTRY      0
```
axel-h, line 74: This should get a more verbose comment. How much of a device tree standard is this actually or is it just what everybody seems to use. I think it might be better to inline these values and not expose any constants. Then a comment there can explain the the magic values come from.  We can factor this out when these constants are to be somewhere. If we provide defines here, we names must be much more specific to avoid conflicts or get confusing.

chrisguikema: I found a link with the standard. Its not just what everyone seems to use.

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -58,9 +58,65 @@
...
+        return -1;
+    }
+
+    c = fdt_get_property(vm->fdt_ori, gic_offset, "reg", &len);
+    if (!c) {
+        ZF_LOGE("Failed to find reg in gic node");
+        return -1;
+    }
+
+    int calculated_size = size_cells * address_cells * NUM_GIC_REGS * sizeof(uint32_t);
```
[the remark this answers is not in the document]
chrisguikema: I updated the order and added a comment. I like keeping the `NUM_GIC_REGS` because its more descriptive than just writing 4. And to virtualize a GIC, you need the 4 entries, so we aren't at risk of not supporting anything here.

libsel4vm/src/arch/arm/vgic/vgic_v2.c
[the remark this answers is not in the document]
kent-mcleod: My preference would be to avoid adding the globals in this PR. The vgic_dist_device struct is intended for storing these exact mapping addresses and so they should be added there. As they aren't the same for the gic_v3, it's fine to create a different vgic_dist_device that's specialized to the gicv2 as it's private to this compilation unit.

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -236,6 +297,9 @@ static vm_frame_t vgic_vcpu_iterator(uintptr_t addr, void *cookie)
  */
 int vm_install_vgic(vm_t *vm)
 {
+    /* Read device tree to find memory regions */
+    gic_get_registers_from_fdt(vm, &dist_addr, &cpu_addr, &vcpu_addr);
```
kent-mcleod, line 301: We should probably check and return the error here if the GIC can't be found in the FDT.

libsel4vm/include/sel4vm/guest_vm.h
```diff
@@ -203,6 +203,9 @@ struct vm {
     char *vm_name;
     unsigned int vm_id;
     bool vm_initialised;
+
+    char *gic_path;
+    void *fdt_ori;
```
kent-mcleod, line 208: As these are both Arm specific, they should be added to the vm_arch struct.

kent-mcleod:
> This commit adds an "fdt_ori" and "gic_node" parameter to the vm struct,

This isn't an insignificant design change to the vm_t struct. If this is something that we should add, then it probably needs to be properly integrated into the rest of the library with proper initialization functions. If it's just an initial way to prototype associating a device tree with a guest for virtual devices to more easily find configuration information it should probably still have a proper way of installing it. (This would also allow the camkes-vm layer to more explicitly generate the fdt and then install it into the guest for the rest of the guest devices to use). Also, how is this expected to work with guests that don't use an FDT but want to use virtual devices that do use an fdt?

kent-mcleod:
One more comment, if this isn't something that we want to tackle now, then maybe a smaller-impact change would be choosing to have device specific interfaces for vm_install_vgic() that could take the configuration arguments needed, allowing the caller (camkes-vm) to do the device tree parsing and pass in the address ranges via the newer vm_install_vgic() interface.

chrisguikema:
Thanks for the comments, Kent. This was something that I just put together based on the comments on using the GIC path to find the phandle. Basically, its more of a proof of concept. If you want to separate this commit out into a draft, so we can figure out the best way to do this properly while merging in the gic phandle, I'm okay with that approach.

chrisguikema:
> Also, how is this expected to work with guests that don't use an FDT but want to use virtual devices that do use an fdt?

Doesn't the kernel pass an FDT to each VMM regardless of whether the guest needs to use the FDT? So I guess there's a disconnect between the guest's device tree, and the FDT used to describe the hardware resources of the system.

Would it be possible to just use the ps_io_ops interface to get the FDT? Then search for the node with the `arm,gic-400` string?

kent-mcleod:
> Doesn't the kernel pass an FDT to each VMM regardless of whether the guest needs to use the FDT? So I guess there's a disconnect between the guest's device tree, and the FDT used to describe the hardware resources of the system.

In CAmkES, the DTB is given to the component if it has the `dtb` property set in its configuration, so it's not always expected to be present.  It would be reasonable for a VM component to load a file from the fileserver component and not need the system DTB given to the component at all.

I think one question that needs to be decided is whether device trees are a concept that libsel4vm should know about and have a mechanism for managing a device tree and installing it into the guest. It seems to me that this is better suited for libsel4vmmplatsupport?

kent-mcleod:
> Thanks for the comments, Kent. This was something that I just put together based on the comments on using the GIC path to find the phandle

Yea, I'm not trying to say you shouldn't have proposed the changes, just pointing out that adding a way for vm devices to query a device tree associated with the guest is something that would be quite useful to many more devices than just the vgic, so we should consider the wider implications.  EG, should this device tree be the system device tree, or the one that the guest sees? Are devices allowed to mutate it? If it's using a libfdt compatible object, then we could allow devices to install nodes in during their own initialization and have an explicit finalization step which is what Qemu does.

libsel4vm/src/arch/arm/vgic/vgicv2_defs.h
axel-h: Can you rename this to e.g. `NUM_FDT_GIC_REGS` and `FDT_GIC_REG_IDX_<DIST|CPU|VCPU_CNTR|VCPU>`? So we have names that are a bit less ambiguous and clearly state this is device tree related

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -58,9 +58,70 @@
...
+    }
+
+    int parent_offset = fdt_parent_offset(vm->fdt_ori, gic_offset);
+    if (0 > parent_offset) {
+        ZF_LOGE("Failed to find gic parent offset");
+        return -1;
+    }
+
+    int address_cells = fdt_address_cells(vm->fdt_ori, parent_offset);
+    int size_cells = fdt_address_cells(vm->fdt_ori, parent_offset);
```
axel-h, line 86:
```suggestion
    int size_cells = fdt_size_cells(vm->fdt_ori, parent_offset);
```

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -58,9 +58,70 @@
...
+
+    /*
+     * 1st reg entry: Distributor
+     * 2nd reg entry: CPU Interface
+     * 3rd reg entry: GIC Virtual Interface Control Register
+     * 4th reg entry: GIC Virtual CPU Interface Register
+     */
+    uint64_t gic_regs[NUM_GIC_REGS];
+    int calculated_size = size_cells * address_cells * NUM_GIC_REGS * sizeof(uint32_t);
+    int reg_offset = calculated_size / NUM_GIC_REGS;
```
axel-h, line 107:
```suggestion
    int reg_size = calculated_size / NUM_GIC_REGS;
```
This is the size of one value, not the offset

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -58,9 +58,70 @@
...
+
+    if (len != calculated_size) {
+        ZF_LOGE("Register size is unexpected. %d != %d", len, calculated_size);
+        return -1;
+    }
+
+    for (int i = 0; i < NUM_GIC_REGS; i++) {
+        memcpy(&gic_regs[i], (void *)(c->data + (i * reg_offset)), sizeof(uint32_t) * address_cells);
```
axel-h, line 115: I think this has an implicit assumption that `address_cells` is 2 and address_cells is 1. That may hold practically, so we could simply reject anything else for now.

chrisguikema: This should work with any case of address/size being 1 or 2.

libsel4vm/src/arch/arm/vgic/vgic_v2.c
```diff
@@ -58,9 +58,70 @@
...
+    }
+
+    /*
+     * 1st reg entry: Distributor
+     * 2nd reg entry: CPU Interface
+     * 3rd reg entry: GIC Virtual Interface Control Register
+     * 4th reg entry: GIC Virtual CPU Interface Register
+     */
+    uint64_t gic_regs[NUM_GIC_REGS];
+    int calculated_size = size_cells * address_cells * NUM_GIC_REGS * sizeof(uint32_t);
```
axel-h, line 106:
```suggestion
    int calculated_size = (address_cells + size_cells)  * NUM_GIC_REGS * sizeof(uint32_t);
```
We have one address cell followed by a size cell for each element. This just work because `2 * 2` happens to equal `2 + 2`

libsel4vm/src/arch/arm/vgic/vgic_v2.c
axel-h: Right, extracting the value works for 1 or 2. But we need `uint64_t gic_regs[NUM_GIC_REGS] = {0};} then to ensure the array is filled with zeros. Otherwise we might just copy 32 bits and the other 32-bit are random data,

chrisguikema:
I'm going to mark this as a draft, since like Kent mentioned, in order to do this, you need access to the FDT in the libraries, which is a bigger design choice than just this PR. I think this would still be useful as a reference for getting the GIC information out of the device tree once the FDT is available.

PR #84, opened 2022-11-11
Arm VM Multicore Improvements

Apave24:
These commits allow for more flexible core configurations for Arm VM systems


libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,23 +50,37 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
...
+            /* Assign vcpu to physical cpu specified in config */
             if (is_vcpu_online(target_vcpu)) {
                 smc_set_return_value(&regs, PSCI_ALREADY_ON);
+            } else if (start_new_vcpu(target_vcpu, entry_point_address, context_id, target_vcpu->target_cpu) == 0) {
+                smc_set_return_value(&regs, PSCI_SUCCESS);
             } else {
                 smc_set_return_value(&regs, PSCI_INTERNAL_FAILURE);
             }
+        } else {
+            ZF_LOGE("Target CPU was an invalid number");
```
axel-h, line 82: We should also log the invalid number
```suggestion
            ZF_LOGE("Target CPU is invalid: %d", target_vcpu->target_cpu);
```

libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,23 +50,37 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
...
+            break;
+        }
+
+        /* Automatically assign vcpu to an unassigned physical cpu */
+        if (target_vcpu->target_cpu == -1) {
+            int selected_cpu = vm_find_free_unassigned_cpu(vcpu->vm);
+            if ((selected_cpu >= 0) && start_new_vcpu(target_vcpu, entry_point_address, context_id, selected_cpu) == 0) {
                 smc_set_return_value(&regs, PSCI_SUCCESS);
             } else {
                 smc_set_return_value(&regs, PSCI_INTERNAL_FAILURE);
```
axel-h, line 70: Let's add a message here also now that some auto-magic is happening here
```
ZF_LOGE("No unused physical core left for vCPU %u", vcpu->vcpu_id);
```

libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,23 +50,37 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
...
             }
-        } else {
+        } else if ((target_vcpu->target_cpu >= 0) && (target_vcpu->target_cpu < CONFIG_MAX_NUM_NODES)) {
+            /* Assign vcpu to physical cpu specified in config */
             if (is_vcpu_online(target_vcpu)) {
                 smc_set_return_value(&regs, PSCI_ALREADY_ON);
+            } else if (start_new_vcpu(target_vcpu, entry_point_address, context_id, target_vcpu->target_cpu) == 0) {
+                smc_set_return_value(&regs, PSCI_SUCCESS);
```
axel-h, line 77: Maybe log an error here also?
ZF_LOGE("[vCPU %u] could not start physical CPU %d , target_vcpu->target_cpu);

libsel4vm/include/sel4vm/guest_vm_util.h
```diff
@@ -72,6 +72,30 @@ static inline vm_vcpu_t *vm_find_free_unassigned_vcpu(vm_t *vm)
...
 
+/***
+ * @function vm_find_free_unassigned_cpu(vm)
+ * Find an unallocated physical cpu
+ * @param {vm_t *} vm           A handle to the vm owning the vcpu
+ * @return                      -1 if no pcpu can be found, otherwise the available pcpu
+ */
+static inline int vm_find_free_unassigned_cpu(vm_t *vm)
+{
+    for (int core = 0; core < vm->num_vcpus; core++) {
```
axel-h, line 83: I'm not sure I get this algorithm, shouldn't we loop over the physical cores in the outer loop and check in the innter loop if this core is used by any VM?

libsel4vm/include/sel4vm/guest_vm_util.h
```diff
@@ -72,6 +72,30 @@ static inline vm_vcpu_t *vm_find_free_unassigned_vcpu(vm_t *vm)
...
+        int core_allocated = 0;
+        for (int i = 0; i < vm->num_vcpus; i++) {
+            if (vm->vcpus[i]->target_cpu == core) {
+                core_allocated = 1;
+                break;
+            }
+        }
+        if (core_allocated) {
```
axel-h, line 91: reverse the ceck, then there is no need for the explicit `continue` and the algorithm is a bit easier to read
```suggestion
        if (!core_allocated) { 
            return core
        }
```

libsel4vm/src/boot.c
```diff
@@ -88,10 +88,17 @@ int vm_assign_vcpu_target(vm_vcpu_t *vcpu, int target_cpu)
...
     }
     vm_vcpu_t *target_vcpu = vm_vcpu_for_target_cpu(vcpu->vm, target_cpu);
-    if (target_vcpu) {
+    if (target_vcpu && (target_vcpu != vcpu)) {
         ZF_LOGE("Failed to assign target cpu - A VCPU is already assigned to core %d", target_cpu);
         return -1;
     }
+
+#if CONFIG_MAX_NUM_NODES > 1
+    if (seL4_TCB_SetAffinity(vcpu->tcb.tcb.cptr, target_cpu)) {
```
axel-h, line 97: Nitpicking: this just moves the check, but the logic is remains semantically broken here. It's easier to understand if this would be written as: 

```suggestion
    if (0 != seL4_TCB_SetAffinity(vcpu->tcb.tcb.cptr, target_cpu)) {
```
Or even more verbosely with an error mesage:
```
    int err = seL4_TCB_SetAffinity(vcpu->tcb.tcb.cptr, target_cpu);
    if (err) {
        ZF_LOGE("[vCPU %u] Failed to set CPU affinity to %d", target_cpu);
   }
```

libsel4vm/include/sel4vm/guest_vm_util.h
Apave24: Agreed, CONFIG_MAX_NUM_NODES probably makes more sense here.

libsel4vm/include/sel4vm/guest_vm_util.h
Apave24: Done

libsel4vm/src/boot.c
Apave24: Done

libsel4vmmplatsupport/src/arch/arm/psci.c
Apave24: Done

libsel4vmmplatsupport/src/arch/arm/psci.c
Apave24: Done in the else case

libsel4vmmplatsupport/src/arch/arm/psci.c
Apave24: Done

libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,23 +50,39 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
...
-        uintptr_t target_cpu = smc_get_arg(&regs, 1);
+        uintptr_t requested_cpu = smc_get_arg(&regs, 1);
         uintptr_t entry_point_address = smc_get_arg(&regs, 2);
         uintptr_t context_id = smc_get_arg(&regs, 3);
-        vm_vcpu_t *target_vcpu = vm_vcpu_for_target_cpu(vcpu->vm, target_cpu);
-        if (target_vcpu == NULL) {
-            target_vcpu = vm_find_free_unassigned_vcpu(vcpu->vm);
-            if (target_vcpu && start_new_vcpu(target_vcpu, entry_point_address, context_id, target_cpu) == 0) {
+        vm_vcpu_t *target_vcpu = NULL;
+        if ((requested_cpu >= 0) && (requested_cpu < vcpu->vm->num_vcpus)) {
```
axel-h, line 57: The `uintptr_t` can never be negative, so we just need to check the upper limit.

libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,23 +50,39 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
...
+            break;
+        }
+
+        /* Automatically assign vcpu to an unassigned physical cpu */
+        if (target_vcpu->target_cpu == -1) {
+            int selected_cpu = vm_find_free_unassigned_cpu(vcpu->vm);
+            if ((selected_cpu >= 0) && start_new_vcpu(target_vcpu, entry_point_address, context_id, selected_cpu) == 0) {
                 smc_set_return_value(&regs, PSCI_SUCCESS);
             } else {
+                ZF_LOGE("[vCPU %u] no unused physical core left", vcpu->vcpu_id);
```
axel-h, line 70: ...or starting failed. I think separating the errors is worth it, because that make debugging easier when something fails here or got out if sync somehow.

libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,15 +50,24 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
         smc_set_return_value(&regs, 0x00010000); /* version 1 */
         break;
     case PSCI_CPU_ON: {
-        uintptr_t target_cpu = smc_get_arg(&regs, 1);
+        uintptr_t requested_cpu = smc_get_arg(&regs, 1);
```
axel-h, line 53: I think the guest would still pass MPIDR here, and that is not a linear number if there are multiple clusters to emulate. So we cannot use this as array index, but have to look it up in all vCPUs the guest got. We should really call even the variable `target_vmpidr` then.

libsel4vm/src/boot.c
```diff
@@ -92,6 +92,15 @@ int vm_assign_vcpu_target(vm_vcpu_t *vcpu, int target_cpu)
         ZF_LOGE("Failed to assign target cpu - A VCPU is already assigned to core %d", target_cpu);
         return -1;
     }
+
+#if CONFIG_MAX_NUM_NODES > 1
+    int err = seL4_TCB_SetAffinity(vcpu->tcb.tcb.cptr, target_cpu);
```
axel-h, line 97: I think to commit is ok. But it makes this generic code now, how do other architectures cope with this change then? Do we gave to change something in the x86 VMM part also?

axel-h:
Since this PR has multiple commits that seem independent, but have dependencies on other PRs, could you split this up that we can merge things independently and avoid unnecessary blocking?

libsel4vm/src/boot.c
Apave24: I believe `vm_assign_vcpu_target` is currently only used by ARM

Apave24:
> Since this PR has multiple commits that seem independent, but have dependencies on other PRs, could you split this up that we can merge things independently and avoid unnecessary blocking?

Dropped commit `libsel4vm: guest cpu requests to vcpus from pcpus` from this PR and added to #85

Apave24:
Bumping this PR. I believe everything has been addressed.

axel-h:
@Apave24 could you rebase this, then we also get another CI run there to see if this passes.

chrisguikema:
@axel-h @abrandnewusername @kent-mcleod 

Hey guys, just wanted to ping on this PR. This series of PRs makes it so that when the VM kicks off a secondary core, the VMM pins the VCPU to a PCPU defined in the `vmX.pcpus` list. I feel like this behavior makes the most sense given the current VMM configuration. Is this a good path forward? Can we work on getting this merged?

PR #85, opened 2022-11-17
libsel4vm: guest cpu requests to vcpus from pcpus

Apave24:
Originally the guests were able to make requests for a specific physical core based on its id. These changes associate that requested id with a vcpu instead of a pcpu.

Signed-off-by: Alex Pavey <Alex.Pavey@dornerworks.com>

### DIFF
--- a/libsel4vmmplatsupport/src/arch/arm/psci.c
+++ b/libsel4vmmplatsupport/src/arch/arm/psci.c
@@ -50,15 +50,24 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
         smc_set_return_value(&regs, 0x00010000); /* version 1 */
         break;
     case PSCI_CPU_ON: {
-        uintptr_t target_cpu = smc_get_arg(&regs, 1);
+        uintptr_t requested_cpu = smc_get_arg(&regs, 1);
         uintptr_t entry_point_address = smc_get_arg(&regs, 2);
         uintptr_t context_id = smc_get_arg(&regs, 3);
-        vm_vcpu_t *target_vcpu = vm_vcpu_for_target_cpu(vcpu->vm, target_cpu);
-        if (target_vcpu == NULL) {
-            target_vcpu = vm_find_free_unassigned_vcpu(vcpu->vm);
-            if (target_vcpu && start_new_vcpu(target_vcpu, entry_point_address, context_id, target_cpu) == 0) {
+        vm_vcpu_t *target_vcpu = NULL;
+        if ((requested_cpu >= 0) && (requested_cpu < vcpu->vm->num_vcpus)) {
+            target_vcpu = vcpu->vm->vcpus[requested_cpu];
+        } else {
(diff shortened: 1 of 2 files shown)


libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,15 +50,24 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
         smc_set_return_value(&regs, 0x00010000); /* version 1 */
         break;
     case PSCI_CPU_ON: {
-        uintptr_t target_cpu = smc_get_arg(&regs, 1);
+        uintptr_t requested_cpu = smc_get_arg(&regs, 1);
```
axel-h, line 53: so, my question holds - we get a (v)MPIDR here from the guest and not a linear (seL4 affinity) ID?

libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,15 +50,24 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
...
         uintptr_t entry_point_address = smc_get_arg(&regs, 2);
         uintptr_t context_id = smc_get_arg(&regs, 3);
-        vm_vcpu_t *target_vcpu = vm_vcpu_for_target_cpu(vcpu->vm, target_cpu);
-        if (target_vcpu == NULL) {
-            target_vcpu = vm_find_free_unassigned_vcpu(vcpu->vm);
-            if (target_vcpu && start_new_vcpu(target_vcpu, entry_point_address, context_id, target_cpu) == 0) {
+        vm_vcpu_t *target_vcpu = NULL;
+        if ((requested_cpu >= 0) && (requested_cpu < vcpu->vm->num_vcpus)) {
+            target_vcpu = vcpu->vm->vcpus[requested_cpu];
```
axel-h, line 58: This will not work, because `requested_cpu` is a MPID value now - because that is what a caller passed into the SMC according to the spec. So we need to do a lookup over all our vCPUs.

libsel4vmmplatsupport/src/arch/arm/psci.c
```diff
@@ -50,15 +50,24 @@ int handle_psci(vm_vcpu_t *vcpu, seL4_Word fn_number, bool convention)
...
+        } else {
+            smc_set_return_value(&regs, PSCI_INTERNAL_FAILURE);
+            break;
+        }
+
+        /* Automatically assign vcpu to an unassigned physical cpu */
+        if (target_vcpu->target_cpu == -1) {
+            int selected_cpu = vm_find_free_unassigned_cpu(vcpu->vm);
+            if ((selected_cpu >= 0) && start_new_vcpu(target_vcpu, entry_point_address, context_id, selected_cpu) == 0) {
```
axel-h, line 67: Don't we have to set the MPIDR of the vCPU here also now, so this is in sync with what the caller requested? Can the caller actually starts arbitrary cored here - shouldn't we have prepared the vCPU list in advance, about what cores the caller can start, ie. what cores the VM has.

PR #86, opened 2022-11-18
x86 interrupt improvements

andybui01:
This PR attempts to do the following:
- Patch up flaky ACPI table support
- Complete the APIC emulation code. In its current state the emulated APIC is not completed and used.
- Add support for Message Signaled Interrupts (MSIs). This allows a greater range of devices to be supported the need for IOAPIC emulation (a bit harder to set up due to the DSDT and AML bytecode. Although I think dornerworks have an attempt at this?)
- WIP: APIC timer support, mostly works just probabilistic and freezes up here and there.

A working manifest for the supermicro board can be found here: https://github.com/andybui01/camkes-vm-examples-manifest/blob/andyb/supermicro_msi/master.xml


chrisguikema:
@andybui01 glad to see someone else has worked on this and released a version for the community! 

I noticed you mentioned DornerWorks in the PR, and you're right that we did some work on this topic, but our code can't be released yet :( What i've noticed is that MSIs don't need the IOAPIC for functionality, which is good, because adding IOAPIC support is a _pain_, due to needed to compile your own DSDT which needs to match the definitions in the `.camkes` file.

andybui01:
> @andybui01 can you explain the non-deterministic behavior you're seeing with the lapic timer? I pulled this into a code base that is running real-time Linux, and the timer seemed to initialize properly, but then stop triggering during the boot process. I'm unsure of whether that is Linux stopping the timer for some reason, or the behavior you noticed.

Hi, thanks for having a look. It's been a while since I worked on this, I remember it would work sometimes and then a fresh boot would freeze things up. I never investigated beyond this.

chrisguikema:
I looked into this a bit, and on my boot, Linux is switching to oneshot mode, and the time server is never set to trigger an event.

chrisguikema:
@andybui01 quick question - where was the reference for the lapic timer? KVM?

andybui01:
> @andybui01 quick question - where was the reference for the lapic timer? KVM?

Yep, KVM.

PR #89, opened 2023-01-11
trivial: revert Copyright headers

abrandnewusername:
Fix the Copyright headers that have been altered incorrectly in a previous [PR](https://github.com/seL4/seL4_projects_libs/pull/74#issuecomment-1328653368).


lsf37 (approved):
Looks good. It's Ok to add a new UNSW line if there were content changes in the file (just not delete old ones).

PR #91, opened 2023-03-21
libsel4vmmplatsupport: remove superfluous statement

axel-h:
The last return statement is never reached.

Note that I think the `ram_base + 0x8000` might be wrong also and should also become  `vm->entry` now. But I will address this in a later PR then after more testing (https://github.com/Hensoldt-Cyber/seL4_projects_libs/commits/patch-axel-26a). It seems that zImage is no longer used for 64-bit anyway.


kent-mcleod (approved):

kent-mcleod:
> It seems that zImage is no longer used for 64-bit anyway.

AFAIK zimage is only an arm32 format. 
It might be preferable to use the equivalent compressed image format for 64bit too in the future.

PR #93, opened 2023-03-21
libsel4vmmplatsupport: use uintptr_t

axel-h:
- This is no pointer that can be dereferenced.
- uintptr_t is used in other places already, so casting can be avoided.


kent-mcleod (approved):
This looks like a good improvement

axel-h:
Can we merge this?

lsf37 (approved):

PR #94, opened 2023-03-21
libsel4vmmplatsupport: use vm->entry for zImage also

axel-h:
This is a follow-up from the change introduced by https://github.com/seL4/seL4_projects_libs/pull/81

- Remove hard coded fall-back address for zImage also.
- inline `zImage_get_load_address()`, so things get a bit simpler .
- Add comment about zImage usage in Linux. Seems it's 32-bit only anyway.


axel-h:
To me, the `vm->entry` usage seems quite strange. Loading behavior of the kernel depends on the image type:
- for a binary, the parameter `load_base_addr` is ignores and the blob is loaded `vm->entry`. That seem the only way how things work in AARCH64
- a zImage (which practically exists for AARCH32 only), `load_base_addr` is also ignored and the address from the header is used. Or, if this is zero, it falls back to `vm->entry` with this PR (before it was indeed `load_base_addr + offset` then.),

So it seems `vm->entry` is slightly misused in what we have currently? Looks like there should be a `vm->kernel_addr` also that must be given in case the kernel is a raw binary.

axel-h:
With the changes of this PR it seems we don't really need 'vm.entry' that 58dc2820 recently added (https://github.com/seL4/seL4_projects_libs/pull/81). We should instead pass the value in `load_base_addr`, as this parameter has been practically unused. That would nicely simplify things then.

PR #100, opened 2023-05-15
Cross-connector: use dataport frame size instead of 4k default

hlyytine:
The main reason for this change set is to support 2 MB and bigger CAmkES dataports. The tooling generates as large as possible frames for dataports, and for example on aarch64 the next step is 2 MB large page. However, the code deals only with 4 kB small pages. Let's use dataport's frame size instead of that default.

### DIFF
--- a/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
+++ b/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
@@ -301,8 +301,6 @@ static int initialise_connections(vm_t *vm, uintptr_t connection_base_addr, cros
 int cross_vm_connections_init_common(vm_t *vm, uintptr_t connection_base_addr, crossvm_handle_t *connections,
                                      int num_connections, vmm_pci_space_t *pci, alloc_free_interrupt_fn alloc_irq)
 {
-    uintptr_t guest_paddr = 0;
-    size_t guest_size = 0;
     if (num_connections > MAX_NUM_CONNECTIONS) {
         ZF_LOGE("Unable to register more than %d dataports", MAX_NUM_CONNECTIONS);
         return -1;
--- a/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
+++ b/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
@@ -48,8 +48,8 @@ struct dataport_iterator_cookie {
     vm_t *vm;
 };
 
-struct connection_info info[MAX_NUM_CONNECTIONS];
-int total_connections;
+static struct connection_info info[MAX_NUM_CONNECTIONS];
+static int total_connections;
 
 static int construct_connection_bar(vm_t *vm, struct connection_info *info, int num_connections, vmm_pci_space_t *pci)
 {
--- a/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
+++ b/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
@@ -188,6 +188,12 @@ static int reserve_dataport_memory(vm_t *vm, crossvm_dataport_handle_t *dataport
     vm_memory_reservation_t *dataport_reservation = vm_reserve_memory_at(vm, dataport_address, size,
                                                                          default_error_fault_callback,
                                                                          NULL);
+    if (!dataport_reservation) {
+        ZF_LOGE("Cannot reserve %zu bytes at 0x%"PRIxPTR, size,
+                dataport_address);
+        return -1;
+    }
+
     struct dataport_iterator_cookie *dataport_cookie = malloc(sizeof(struct dataport_iterator_cookie));
     if (!dataport_cookie) {
         ZF_LOGE("Failed to allocate dataport iterator cookie");
--- a/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
+++ b/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
@@ -41,13 +41,6 @@ struct connection_info {
     int connection_irq;
 };
 
-struct dataport_iterator_cookie {
-    seL4_CPtr *dataport_frames;
-    uintptr_t dataport_start;
-    size_t dataport_size;
-    vm_t *vm;
-};
-
 static struct connection_info info[MAX_NUM_CONNECTIONS];
 static int total_connections;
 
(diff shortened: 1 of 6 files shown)


libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
```diff
@@ -153,57 +153,67 @@ static int reserve_event_bar(vm_t *vm, uintptr_t event_bar_address, struct conne
...
+    struct dataport_iterator_cookie *dataport = cookie;
+
+    uintptr_t frame_start = ROUND_DOWN(addr, BIT(dataport->frame_size_bits));
+    if (frame_start < dataport->base) {
+        ZF_LOGE("Access outside dataport");
+        return frame_result;
+    }
+
+    size_t offset = frame_start - dataport->base;
+    if (!IS_ALIGNED(offset, dataport->frame_size_bits)) {
```
axel-h, line 166: Maybe this should get some comments what calling contract we are expecting here, so it's more clear where the are checking violations of the calling contract vs checking bad configurations that can still happen as they are not in the scope of the calling contract. Also, we make the error messages a bit more verbose about what address exactly failed here - then debugging gets a bit easier with the addresses printed.
- It seem we expect proper alignment of `dataport->base`
- Also, `addr` is always in the range "`dataport->base` to  `dataport->num_frames * dataport->frame_size_bits`"
- We do not expect proper alignment of `addr`. But there seems there is a guarantee/expectation that the iterating loop that calls us it incrementing `addr` by what we return in `frame_result.size_bits`. So, can `addr` ever be misaligned except in the first rounds?

axel-h:
Nice work.

- I'd squash the commit "cross_vm_connectors: Remove redundant field" into "cross_vm_connectors: Use dataport frame size", so we never have a redundant field. This suashing makes the commit become the compatibility breaker, but it also shows more clearly why we did it.
- I could not find anything in the history that shows why `dataport_iterator_cookie` had a VM reference. Does anybody else know?

kent-mcleod:
> Test with: [seL4/camkes-tool#134](https://github.com/seL4/camkes-tool/pull/134)

I didn't see where this PR depended on the changes in the other PR?

libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
hlyytine: Rearranged code so that `base` alignment is checked against the frame bit size. After that we have only one sanity check that `frames` inside cookie is not accessed out of bounds. But if that happens, something has gone south big time, as it seems `vm_map_reservation()` is not doing what we told it to do. Hence I feel something more heavyweight debug tool than just a print here is needed, hence I would choose not to make error message more verbose (also I couldn't fit it in 80 columns :D)

hlyytine:
> * I'd squash the commit "cross_vm_connectors: Remove redundant field" into "cross_vm_connectors: Use dataport frame size", so we never have a redundant field. This suashing makes the commit become the compatibility breaker, but it also shows more clearly why we did it.

Yes I agree, done now.

hlyytine:
> > Test with: [seL4/camkes-tool#134](https://github.com/seL4/camkes-tool/pull/134)
> 
> I didn't see where this PR depended on the changes in the other PR?

Indirectly via seL4/camkes-vm#97, I listed both here, @axel-h removed camkes-vm from here.

hlyytine:
Sorry folks, pushed some unrelated content by accident, and now convenience links do not show anything sensible, here is a comparison link for your convenience: [compare](urlhttps://github.com/seL4/seL4_projects_libs/compare/3975c237afe9777d9100712b910b4c24bceb952e..d1007b6e59269a2f8b3e25b478d30e237a906c4d)

kent-mcleod:
> Indirectly via [seL4/camkes-vm#97](https://github.com/seL4/camkes-vm/pull/97), I listed both here, @axel-h removed camkes-vm from here.

I see.  Currently it doesn't appear that these three PRs would work together because none of the connectors templates have been updated to implement `dataport_get_page_size_bits_fn`?

hlyytine:
> I see. Currently it doesn't appear that these three PRs would work together because none of the connectors templates have been updated to implement `dataport_get_page_size_bits_fn`?

That is easy explain: I just forgot to create PR like this: seL4/global-components#48

Added the PR to "test with" field in `camkes-vm` but it appears somebody needs to nudge CI to get it to run.

libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
```diff
@@ -151,59 +151,68 @@ static int reserve_event_bar(vm_t *vm, uintptr_t event_bar_address, struct conne
     return 0;
 }
 
-static vm_frame_t dataport_memory_iterator(uintptr_t addr, void *cookie)
+static vm_frame_t dataport_memory_iterator(uintptr_t frame_start, void *cookie)
```
hlyytine, line 154: Based on discussion and my tentative conclusion in seL4/global-components#48, isn't a more proper name for the first argument `page_start`? In this function / file and everywhere else?

libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
```diff
@@ -151,59 +151,68 @@ static int reserve_event_bar(vm_t *vm, uintptr_t event_bar_address, struct conne
...
-        frame_start > dataport_start + dataport_size) {
-        ZF_LOGE("Error: Not Dataport region");
+    struct dataport_iterator_cookie *dataport = cookie;
+    size_t frame_size_bits = dataport->frame_size_bits;
+
+    size_t frame_idx = (frame_start - dataport->base) >> frame_size_bits;
+    if (frame_idx >= dataport->num_frames) {
+        ZF_LOGE("Attempt to map outside dataport");
         return frame_result;
     }
```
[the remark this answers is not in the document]
hlyytine: To answer my own question, I added two commits that a) commonize mapping an array of frames into reservation (they are not dataport specific at all) and b) move the commonized code to `libsel4vm`. This way this driver-like code does not contain any internals of memory mapping code. IMO it solves the question here: `vm_map_frames()` will sanitize the values and the iterator function gets only the sane values. We still do rudimentary error check there, but we don't intend to give verbose explanation in the error message.

libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
hlyytine: Used `page_start` in this PR.

libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
hlyytine: > Maybe this should get some comments what calling contract we are expecting here, so it's more clear where the are checking violations of the calling contract vs checking bad configurations that can still happen as they are not in the scope of the calling contract. 

I realized lots of my additional checks changed the calling contract and removed those checks. My intent was to prevent anyone shooting their feet but perhaps sometimes it gets in the way. So I am now adding only the essentials.

> Also, we make the error messages a bit more verbose about what address exactly failed here - then debugging gets a bit easier with the addresses printed.

The only one added check -- an error check and not calling contract enforcement -- has now the details.

> * It seem we expect proper alignment of `dataport->base`
> * Also, `addr` is always in the range "`dataport->base` to  `dataport->num_frames * dataport->frame_size_bits`"
> * We do not expect proper alignment of `addr`. But there seems there is a guarantee/expectation that the iterating loop that calls us it incrementing `addr` by what we return in `frame_result.size_bits`. So, can `addr` ever be misaligned except in the first rounds?

These questions are now solved by mimicking the style of other mapping code, just adding enough checks that the code does not access out of bounds etc.

hlyytine:
Refactored the commits; also decided not to touch any external APIs, so this PR is now standalone, not depending on any other PR. I think now I'm not changing behaviour of anything; except of course fixing that broken larger dataports.

kent-mcleod (approved):

kent-mcleod:
> Refactored the commits; also decided not to touch any external APIs, so this PR is now standalone, not depending on any other PR. I think now I'm not changing behaviour of anything; except of course fixing that broken larger dataports. Please check out.

I should have read this before reviewing the camkes-tool and global-components PRs, although I think being able to query the frame size in bits has a small advantage than having to calculate it by dividing the total size by the number of frames and then converting.

kent-mcleod (approved):

kent-mcleod:
Thanks for making these improvements.

PR #106, opened 2023-05-29
Make convenience function publicly available

hlyytine:
Also align with naming and argument order style used in libfdt.

Test with: seL4/camkes-vm#103

### DIFF
--- a/libsel4vmmplatsupport/src/fdt_utils.c
+++ b/libsel4vmmplatsupport/src/fdt_utils.c
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2019, Data61, CSIRO (ABN 41 687 119 230)
+ * Copyright 2023, Technology Innovation Institute
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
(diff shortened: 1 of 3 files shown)


libsel4vmmplatsupport/src/fdt_utils.c
```diff
@@ -0,0 +1,27 @@
+/*
```
axel-h, line 1: I wonder if https://github.com/seL4/util_libs/tree/master/libfdt could be a better place for this. This is generic DTB handling utility code, that would be used outside of sel4vmm also.

hlyytine: Sure it is a better place, but seems that it is an almost vanilla copy of `libfdt`. I was kind of pushed back because I thought it is that way in order to facilitate easier upgrading. Just now I realized there is actually that SPDX commit that makes it not vanilla anymore. I will create a PR in that repo too, so we can discuss which one is correct.

axel-h:
I think we should close this and use https://github.com/seL4/camkes-vm/pull/107 instead.

PR #111, opened 2023-06-15
64-bit-vms: small code refactoring

abrandnewusername:
switch from `CONFIG_ARCH_X86_64` to `CONFIG_X86_64_VTX_64BIT_GUESTS` to allow 32 bit guests on 64 bit hosts.

This commit contains pieces that I omitted in the previous [code refactor](https://github.com/seL4/seL4_projects_libs/pull/16/commits/8227e006f50fbe981a1ce3791da5e11386c65bf5), and should be able to fix the [test failures](https://github.com/seL4/camkes-vm-examples/actions/runs/5273212619/jobs/9536500187).


lsf37 (approved):
I can't tell from the diff if the change is complete (if we're missing more instances), but what is here looks good.

PR #114, opened 2023-09-14
libsel4vm: implement translation vspace

chrisguikema:
This commit keeps the guest memory space mapped into the VMM at all times. This speeds up VMM access to guest memory space, which is vital for high throughput virtio processes.

Furthermore, multicore x86 guests have multiple VMMs running simultaneously. Without the translation vspace, guest memory access must be locked around a mutex, which kills performance.

In order to not break existing functionality, this option is wrapped around a default OFF CMake variable.


chrisguikema:
The translation vspace used to exist before the ARM and x86 libraries were merged. We had to re-implement it in order for multicore x86 guests to be functional. Then when doing virtio-net work, we realized it dramatically increased performance.

PR #115, opened 2023-10-25
sel4vm, lapic: fix apic mmio access

chrisguikema:
The subtraction is backwards, so lapic access will never work.


abrandnewusername (approved):
Thanks!

PR #116, opened 2023-10-25
Guest ram fixes

chrisguikema:
Fixes a couple small bugs we ran into when running 64-bit x86 VMs.

### DIFF
--- a/libsel4vm/src/guest_ram.c
+++ b/libsel4vm/src/guest_ram.c
@@ -48,7 +48,15 @@ static int ram_region_cmp(const void *a, const void *b)
 {
     const vm_ram_region_t *aa = a;
     const vm_ram_region_t *bb = b;
-    return aa->start - bb->start;
+
+    /* qsort expected return values */
+    if (aa->start == bb->start) {
+        return 0;
+    } else if (aa->start < bb->start) {
+        return -1;
+    } else {
+        return 1;
+    }
 }
 
 static void sort_guest_ram_regions(vm_mem_t *guest_memory)
--- a/libsel4vm/src/guest_ram.c
+++ b/libsel4vm/src/guest_ram.c
@@ -185,7 +185,12 @@ int vm_ram_find_largest_free_region(vm_t *vm, uintptr_t *addr, size_t *size)
     /* find the largest unallocated region */
     for (int i = 0; i < guest_memory->num_ram_regions; i++) {
         if (!guest_memory->ram_regions[i].allocated &&
-            guest_memory->ram_regions[i].size > largest_size) {
+            guest_memory->ram_regions[i].size > largest_size
+#ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
+            && ((guest_memory->ram_regions[i].start +
+                 guest_memory->ram_regions[i].size) < UINT32_MAX)
+#endif
+           ) {
             largest = i;
             largest_size = guest_memory->ram_regions[i].size;
         }


libsel4vm/src/guest_ram.c
```diff
@@ -185,7 +192,11 @@ int vm_ram_find_largest_free_region(vm_t *vm, uintptr_t *addr, size_t *size)
     /* find the largest unallocated region */
     for (int i = 0; i < guest_memory->num_ram_regions; i++) {
         if (!guest_memory->ram_regions[i].allocated &&
-            guest_memory->ram_regions[i].size > largest_size) {
+            guest_memory->ram_regions[i].size > largest_size
+#ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
+            && guest_memory->ram_regions[i].start < UINT32_MAX
```
abrandnewusername, line 197: Why does the start address have to be lower than `UINT32_MAX`?

libsel4vm/src/guest_ram.c
```diff
@@ -48,7 +48,14 @@ static int ram_region_cmp(const void *a, const void *b)
...
     const vm_ram_region_t *bb = b;
-    return aa->start - bb->start;
+
+    if (aa->start == bb->start) {
+        return 0;
+    } else if (aa->start < bb->start) {
+        return -1;
+    } else {
+        return 1;
+    }
```
Indanz, line 58: The problem here is that the function returns an `int` while `start` is an `uintptr_t` and the diff between aa and bb may be bigger than 32-bits. So the fix is fine, but the commit message doesn't make much sense to me.

chrisguikema: Updated the commit message to include that information and add that this function is used with qsort.

libsel4vm/src/guest_ram.c
chrisguikema: The structure used to populate the BIOS uses uint32_t.

```
struct setup_header {
    uint8_t    setup_sects;
    uint16_t   root_flags;
    uint32_t   syssize;
    uint16_t   ram_size;
#define RAMDISK_IMAGE_START_MASK    0x07FF
#define RAMDISK_PROMPT_FLAG     0x8000
#define RAMDISK_LOAD_FLAG       0x4000
    uint16_t   vid_mode;
    uint16_t   root_dev;
    uint16_t   boot_flag;
    uint16_t   jump;
    uint32_t   header;
    uint16_t   version;
    uint32_t   realmode_swtch;
    uint16_t   start_sys;
    uint16_t   kernel_version;
    uint8_t    type_of_loader;
    uint8_t    loadflags;
#define LOADED_HIGH (1<<0)
#define QUIET_FLAG  (1<<5)
#define KEEP_SEGMENTS   (1<<6)
#define CAN_USE_HEAP    (1<<7)
    uint16_t   setup_move_size;
    uint32_t   code32_start;
    uint32_t   ramdisk_image;
    uint32_t   ramdisk_size;
```

libsel4vm/src/guest_ram.c
Indanz: Typo: "to sort to guest ram regions." -> "to sort the guest ram regions.", otherwise great. Good to mention qsort.

libsel4vm/src/guest_ram.c
Indanz: But why check the start, surely the end must be `< UINT32_MAX`?

libsel4vm/src/guest_ram.c
axel-h: Thanks. But (nitpicking) I think it would be better to put the qsort usage as comment over the function and not in the commit message then.

libsel4vm/src/guest_ram.c
chrisguikema: The function just checks for the largest free, contiguous region. My guess is that if a region started in 32-bit space, and ended in 64-bit space, that would be fine, since only the start address matters for Linux, and there's only a few hundred bytes of code before the processor state is switched to 64-bit mode.

However, we can check for `.start + .size` if you'd prefer.

libsel4vm/src/guest_ram.c
chrisguikema: I added a comment about qsort return values.

libsel4vm/src/guest_ram.c
Indanz: Well, in that case you should check for `<= UINT32_MAX` instead.

I'm fine with whatever works, but `+ .size` feels more correct to me.

Indanz (approved):

abrandnewusername (approved):

PR #117, opened 2023-10-25
vmmplatsupport: support hpet emulation

chrisguikema:
Real-time Linux kernels often don't work when using the PIT. This commit adds support for emulating the HPET.

### DIFF
--- a/libsel4vmmplatsupport/src/arch/x86/acpi.c
+++ b/libsel4vmmplatsupport/src/arch/x86/acpi.c
@@ -19,8 +19,12 @@ Author: W.A. */
 #include <sel4vm/guest_memory_helpers.h>
 #include <platsupport/plat/acpi/acpi.h>
 
+#include <sel4vmmplatsupport/gen_config.h>
 #include <sel4vmmplatsupport/guest_memory_util.h>
 #include <sel4vmmplatsupport/arch/acpi.h>
+
+#define HPET_DEFAULT_PADDR 0xfed00000
+#define HPET_DEFAULT_TBID  0x8086a201
 
 #define APIC_FLAGS_ENABLED (1)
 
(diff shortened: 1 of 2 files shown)


libsel4vm/src/arch/x86/hpet.c
```diff
@@ -0,0 +1,737 @@
+/*
```
abrandnewusername, line 1: Would it be better to put this file in https://github.com/seL4/camkes-vm/tree/master/components/Init/src along with other device emulators?

libsel4vmmplatsupport/src/arch/x86/acpi.c
```diff
@@ -19,8 +19,10 @@ Author: W.A. */
 #include <sel4vm/guest_memory_helpers.h>
 #include <platsupport/plat/acpi/acpi.h>
 
+#include <sel4vmmplatsupport/gen_config.h>
 #include <sel4vmmplatsupport/guest_memory_util.h>
 #include <sel4vmmplatsupport/arch/acpi.h>
+#include <sel4vmmplatsupport/arch/hpet.h>
```
abrandnewusername, line 25: Maybe add an `#ifdef CONFIG_VMM_USE_HPET` here

libsel4vm/src/arch/x86/hpet.c
chrisguikema: That is where we had it originally, but it got moved to the library. I'm not sure why, but I can try to move it back.

libsel4vmmplatsupport/src/arch/x86/acpi.c
chrisguikema: Sure.

libsel4vm/src/arch/x86/hpet.c
chrisguikema: I moved it back to the Init component.

abrandnewusername (approved):

lsf37 (approved):

PR #118, opened 2023-10-26
vmmplatsupport, x86: remove e820 entry

chrisguikema:
This commit removes an e820 entry for 64-bit guests. This was a reserved memory region from the end of guest memory to the beginning of high memory. When 64-bit guests have access to 64-bit memory, this subtraction would place a bogus region in the e820 map that could confuse the guest.


Indanz (approved):
Assuming it's necessary on 32-bit, otherwise I would just remove it altogether.

abrandnewusername (approved):

chrisguikema:
Can this be moved?

PR #119, opened 2023-10-26
vmmplatsupport, vesa: fix vesa mapping

chrisguikema:
The previous behavior would only attempt to map the vesa framebuffer and set the screen info struct if the vbe protection mode interface was successfully mapped in. This removes the dependency on the vbe protected mode interface. It is not clear that it is needed as vesa framebuffer passthrough works without it.

### DIFF
--- a/libsel4vmmplatsupport/src/arch/x86/guest_boot_init.c
+++ b/libsel4vmmplatsupport/src/arch/x86/guest_boot_init.c
@@ -8,6 +8,7 @@
 
 #include <autoconf.h>
 #include <sel4vm/gen_config.h>
+#include <sel4vmmplatsupport/gen_config.h>
 
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


libsel4vmmplatsupport/src/arch/x86/guest_boot_init.c
```diff
@@ -92,18 +93,19 @@ static void make_guest_screen_info(vm_t *vm, struct screen_info *info)
         }
         if (error) {
             ZF_LOGE("Failed to map vbe protected mode interface for VESA frame buffer. Disabling");
```
abrandnewusername, line 95: The error message need to be coherent with the behavior

abrandnewusername (approved):

libsel4vmmplatsupport/src/arch/x86/guest_boot_init.c
chrisguikema: Switched to "Not Disabling".

PR #120, opened 2023-10-26
vmmplatsupport, pci: fix offset bug

chrisguikema:
The offset is applied twice, which can prevent PCI passthrough from functioning properly when the offset is not zero. This commit removes the second offset addition.


axel-h (approved):
Looks good to me, as the offset is already added in line 146.

lsf37 (approved):

PR #121, opened 2023-11-01
x86, decode: fix 4-level page table walk

chrisguikema:
This commit fixes the page table walk when using 4-level mappings. Linux can use 64-bit memory to store tables, and masking off the upper 32-bits prevents the walk from functioning properly. Additionally, the physical instruction calculation was using the incorrect logical address bits for each level. Finally, the last PTE doesn't have the 7th bit flag a valid page, we just check the first bit to ensure the entry is present, and calculate the PTE address.

### DIFF
--- a/libsel4vm/src/arch/x86/processor/decode.c
+++ b/libsel4vm/src/arch/x86/processor/decode.c
@@ -27,9 +27,17 @@ Author: W.A.
 /* TODO are these defined elsewhere? */
 #define IA32_PDE_SIZE(pde) (pde & BIT(7))
 #define IA32_PDE_PRESENT(pde) (pde & BIT(0))
+
+#ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
+#define IA32_PTE_ADDR(pte) (pte & 0xFFFFFFFFFF000)
+#define IA32_PDPTE_ADDR(pdpte) (pdpte & 0xFFFFFC0000000)
+#define IA32_PDE_ADDR(pde) (pde & 0xFFFFFFFE00000)
+#else
 #define IA32_PTE_ADDR(pte) (pte & 0xFFFFF000)
 #define IA32_PDPTE_ADDR(pdpte) (pdpte & 0xC0000000)
 #define IA32_PDE_ADDR(pde) (pde & 0xFFE00000)
+#endif
+
 #define IA32_PSE_ADDR(pse) (pse & 0xFFC00000)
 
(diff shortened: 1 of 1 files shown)


libsel4vm/src/arch/x86/processor/decode.c
```diff
@@ -27,9 +27,17 @@ Author: W.A.
 /* TODO are these defined elsewhere? */
 #define IA32_PDE_SIZE(pde) (pde & BIT(7))
 #define IA32_PDE_PRESENT(pde) (pde & BIT(0))
+
+#ifdef CONFIG_X86_64_VTX_64BIT_GUESTS
```
Indanz, line 31: Wouldn't it make more sense to always define it like this if the host is 64-bit? The only reason no to have it like that is for 32-it x86, right?

Indanz (approved):
Overall it seems an improvement without breaking anything.

libsel4vm/src/arch/x86/processor/decode.c
chrisguikema: I don't think so. Without `CONFIG_X86_64_VTX_64BIT_GUESTS`, with a 64-bit host you have a 32-bit guest, which cannot have 64-bit memory, so all of the page tables are in low memory where a 32-bit mask is appropriate.

chrisguikema:
Can this be merged?

abrandnewusername (approved):

libsel4vm/src/arch/x86/processor/decode.c
Indanz: But a 32-bit guest can have PAE support too, right?

So not really related to your changes, but who controls what addresses `vm_fetch_instruction()` deals with? That is, are all those assumptions and asserts safe? What if `eip` points to invalid memory?

chrisguikema: No. 32-bit guests do not have PAE.

image

There's also a check to enforce this:

```
    /* ensure that PAE is not enabled */
#ifndef CONFIG_X86_64_VTX_64BIT_GUESTS
    if (cr4 & X86_CR4_PAE) {
        ZF_LOGE("Do not support walking PAE paging structures");
        return -1;
    }
#endif /* not CONFIG_X86_64_VTX_64BIT_GUESTS */
```

Additionally, the macros deal specifically with physical addresses, which in a 32-bit guest will always be 32-bit.

PR #125, opened 2024-02-23
libsel4vm: Remove SMP config guard on vmpidr reg

kent-mcleod:
seL4_ARM_VCPU objects now allow single core configurations to access the VMPIDR register, thus this config guard is no longer needed.


kent-mcleod:
To be merged once https://github.com/seL4/seL4/pull/1181 is

Indanz (approved):

PR #129, opened 2024-11-14
vm_syscall_handler error

kenaml:
err = handle_syscall(vcpu);
If handle_syscall successfully processes, it will return VM_EXIT_HANDLED, which is defined as 1.
if (!err) {
The original condition !err is 0, causing the VM to block after successful processing.


kent-mcleod (approved):
Thanks!

PR #131, opened 2025-03-10
vmmplatsupport: Fix issue with unaligned writes

kent-mcleod:
strncpy was generating unaligned writes if the str length was not alligned. Because the source memory passed into the function's length isn't known, we copy via an alligned stack buffer to avoid an unaligned write to uncached memory.

### DIFF
--- a/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
+++ b/libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
@@ -30,7 +30,7 @@
 #define EVENT_BAR_CONSUME_EVENT_REGISTER 0x4
 #define EVENT_BAR_CONSUME_EVENT_REGISTER_INDEX 1
 #define EVENT_BAR_DEVICE_NAME_REGISTER 0x8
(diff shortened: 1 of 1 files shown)


libsel4vmmplatsupport/src/drivers/cross_vm_connection.c
```diff
@@ -30,7 +30,7 @@
 #define EVENT_BAR_CONSUME_EVENT_REGISTER 0x4
 #define EVENT_BAR_CONSUME_EVENT_REGISTER_INDEX 1
 #define EVENT_BAR_DEVICE_NAME_REGISTER 0x8
-#define EVENT_BAR_DEVICE_NAME_MAX_LEN 50
+#define EVENT_BAR_DEVICE_NAME_MAX_LEN 48
```
kent-mcleod, line 33: TODO: check why this was 50?

lsf37 (approved):

lsf37:
@kent-mcleod  shall we merge this and make an issue for the TODO above?

PR #138, opened 2026-09-18
libvmmplatsupport: bounds checks in virtio_block_emul

lsf37:
- Prevent a potential stack overflow where more than NUM_REQUEST_ADDRS are provided. Explicitly check for NUM_REQUEST_ADDRS and log an error if the number of requests is too high.

- turn `assert` into runtime check: `assert` is compiled out in release mode, which means the check is not performed in production systems. Turn it into a runtime check instead that logs an error and returns to the user without doing anything.


lsf37:
Test running at https://github.com/seL4/camkes-vm-examples/pull/100

Indanz (approved):
The code in this file is a lost cause, it should be burned, it's no where near production ready, I hope no one uses this...

lsf37:
It's pretty rough, yes. The only redeeming thing is that it is usually contained, i.e. fixing things like these in here is just defence in depth.

Overall, the development emphasis is on the Microkit VMM in any case. Once that is all stable and finally on feature parity with CAmkES, we can start the process of deprecating all of this.

